PR #117, opened 2020-09-12
Update SQL script path references.

TurboPT:
There are a few SQL script path references that need the **country_sql** portion of the path specified in the documentation or install text info.

Reported in the forums by Tom Glare:
http://www.weberp.org/forum/showthread.php?tid=8409&pid=16137#pid16137


timschofield:
Thanks Paul.

PR #120, opened 2020-11-15
Master

TurboPT:
Apply forum query contribution provided by **HDeriauFF** to show zero counts within StockUsageGraph periods.

**Paul**: added configurable system parameter to use with the contribution.


huguesderiau:
Great, thank you!

It might be interesting taking a look at https://jpgraph.net for future releases.
I think I will give it a try with my webERP version

PR #132, opened 2022-10-05
Fix GROUP BY qualifiers no longer support by MySQL 8.

TurboPT:
Fix GROUP BY qualifiers removed by MySQL 8, use ORDER BY clause instead.


timschofield:
Thanks Paul

PR #190, opened 2024-10-23
[Upgrade to PHP8.3] IDE-assisted Debugging - casual code scanning - part 1

aingelc12ell:
Fix: Curly brace access syntax as deprecated since PHP 7.4
Fix: Transform each() to foreach() as removed in PHP8.0
Fix: [] operator on strings
Fix: Variables to be passed by reference
Fix: Left and right hand sides of the assignment operator are equal
Fix: Undefined API Error constants and descriptions
Fix: DB_error_no called as string
Fix: String constants as functions
Fix: String constants as quoted strings
Fix: Misplaced break on switch statement
Fix: __construct to __construct()
Fix: Misspelled Ivalid... to Invalid...

### DIFF
--- a/.github/FUNDING.yml
+++ b/.github/FUNDING.yml
@@ -1,0 +1,15 @@
+# These are supported funding model platforms
+
+github: [timschofield]
+patreon: # Replace with a single Patreon username
+open_collective: # Replace with a single Open Collective username
+ko_fi: # Replace with a single Ko-fi username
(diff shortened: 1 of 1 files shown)


timschofield:
Thanks Aingel, just the one comment on all those:

In SuppPriceList.php you are correct to identify the ENT_QUTOES as an error, but it is meant to be the built in constant ENT_QUOTES so shouldn't need quote marks.

Other than that it's all great

Tim

PR #194, opened 2024-10-23
Repushing IDE-assisted debugging and code changes #190 & #193

aingelc12ell:
Fix: Curly brace access syntax as deprecated since PHP 7.4
Fix: Transform each() to foreach() as removed in PHP8.0
Fix: [] operator on strings
Fix: Variables to be passed by reference
Fix: Left and right hand sides of the assignment operator are equal
Fix: Undefined API Error constants and descriptions
Fix: DB_error_no called as string
Fix: String constants as functions
Fix: String constants as quoted strings
Fix: Misplaced break on switch statement
Fix: __construct to __construct()
Fix: Misspelled Ivalid... to Invalid...
Fix: Installer css for default display
Fix: deprecated '${' string interpolation
Fix: Expression result not used within scope
Fix: Improper use of dot to append to variable
Fix: return value on getCharset()
Fix: foreach array is used as value
Fix: Parameter not mapped to conversion specification
Fix: Invalid function result for DB_Tx_Commit, DB_Tx_Rollback, and DB_Txn_Begin
Fix: return value for VerifyPurchDataLineExists()
Fix: prnMsg() can optionally return a value


timschofield:
Thanks Aingel, this has been applied

Tim

PR #196, opened 2024-10-24
IDE-assisted bug fixes - part 3

aingelc12ell:
- Fix: css identifiers and terms
- Fix: Private visibility access via __get()
- Fix: allow strict https implementation be using // instead of http://

### DIFF
--- a/BankAccountUsers.php
+++ b/BankAccountUsers.php
@@ -34,7 +34,7 @@ if (isset($_POST['Cancel'])) {
 
 if (isset($_POST['Process'])) {
 	if ($_POST['SelectedBankAccount'] == '') {
-		echo prnMsg(_('You have not selected any bank account'),'error');
+		prnMsg(_('You have not selected any bank account'),'error');
 		echo '<br />';
 		unset($SelectedBankAccount);
 		unset($_POST['SelectedBankAccount']);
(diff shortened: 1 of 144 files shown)


timschofield:
Thanks for this. It is great, except I'm unsure about the changes to the prnMsg() function. I see you have created the option to print the message "inline" instead of grouping all the meassages at the top. I'm not against this idea (though I would need to see it in action to be sure) but unless I have missed something, wouldn't the message get duplicated, as it is still being added to the $Messages[] array?
prnMsg() used to output inline, but then it was decided it was better to group them together. I think it was @andrewcouling who did the work on this before. Maybe your idea of mixing this is the best idea though.

timschofield:
The more I think about it, the more I like your idea of giving the developer the option of having it echoed inline, or putting it at the top. I think it very much depends on each case. Problem is there are hundreds (thousands??) of instances throughout the whole codebase. I think I would keep your change to the function (just removing the possibility of duplication) but remove all the echoes for now, making a separate project to review the instances of all prnMsg() calls in future. That way we have the facility for developers to choose.
What do you think?

aingelc12ell:
I can make the changes to remove all the echo for now.

And let the new developers choose how they can be implemented.

merge the pull request for now, So I can update my fork from your base

timschofield:
Ok, have done

PR #264, opened 2024-12-01
Added NetBeans IDE project metadata to gitignore

dalers:
OK if we add IDE meta data to .gitignore? If no comments in a day or two I will merge this PR myself (unless someone else does it first)


dalers:
NetBeans gitignore statement should be more general, PR will be closed

PR #265, opened 2024-12-01
Add NetBeans IDE project metadata to gitignore

dalers:
OK to add IDE meta data to .gitignore? If no one disagrees in a day or two I will merge this PR (unless someone else does it first).

Fwiw, I think companies/ subdirectories _except_ for weberpdemo should also be ignored as a convenient safety precaution, but perhaps first "weberpdemo" should be renamed "default" for clarity. I'll post this to Discussions  and if there is support for the proposal I'll investigate doing the work.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,3 @@
 config.php
 # NetBeans project metadata
-/nbproject/private/
+nbproject/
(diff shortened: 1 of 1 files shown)


timschofield:
I have answered the question of th companies/ sub-directory in the discussions

PR #282, opened 2024-12-06
- Fix: Trying to access array offset on value of type null (#251)

aingelc12ell:
Exclude if condition when they are not set by the while loop of records


timschofield:
Thanks for this, I have merged

Tim

PR #295, opened 2024-12-10
Fixes #288 Use config.distrib.php File To Create config.php File With…

### LINKED ISSUE #288: installer created config.php vs project config.distrib.php
### Discussed in https://github.com/timschofield/webERP/discussions/253

<div type='discussions-op-text'>

<sup>Originally posted by **dalers** November 23, 2024</sup>
config.distrib.php has more explanations and additional variables compared to config.php created by the installer, also there seem to be some differences in variables:

- $host in config.distrib.php vs $Host in config.php
- $DefaultDatabase in config.distrib.php vs $DefaultCompany in config.php
- $CompanyList[] array in config.distrib.php vs no list in config.php

I also suspect some variables have been deprecated or are now redundant.

Would it be  possible for config.php to essentially be an edited version of config.distrib.php, so the comments, variables, order, etc. are the same? </div>

bcinteractive:
… Install Data

What: modify code for creating config.php file.
Why: make config.php consistent with config.distrib.php variable-wise.
How: load config.distrib.php and parse each line.  If a line contains a variable that is also found in installer data, modify this line to contain the installer data. Create config.php after parsing all lines.
Note: Must use the same variable names found in config.distrib.php so comparison works.

### DIFF
--- a/install/pages/page_6.php
+++ b/install/pages/page_6.php
@@ -128,68 +128,73 @@ if (!file_exists($Path_To_Root . '/companies/' . $_SESSION['Installer']['Databas
 	}
 }
 
-//$Msg holds the text of the new config.php file
-$Msg = "<?php\n\n";
-$Msg.= "// User configurable variables\n";
-$Msg.= "//---------------------------------------------------\n\n";
-$Msg.= "//DefaultLanguage to use for the login screen and the setup of new users.\n";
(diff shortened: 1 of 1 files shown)


timschofield:
Thanks very much!

Something I can cross off my todo list. Much appreciated

Tim

PR #345, opened 2025-01-06
Fix Updating Default Theme Option In System Parameters

bcinteractive:
The item is located in config.php and the current code doesn't update the config.php for the $DefaultTheme.

Modified the code to properly update config.php with $DefaultTheme.

### DIFF
--- a/SystemParameters.php
+++ b/SystemParameters.php
@@ -124,23 +124,42 @@ if (isset($_POST['submit'])) {
 		if ($_SESSION['DefaultDateFormat'] != $_POST['X_DefaultDateFormat'] ) {
 			$SQL[] = "UPDATE config SET confvalue = '".$_POST['X_DefaultDateFormat']."' WHERE confname = 'DefaultDateFormat'";
 		}
-		if($DefaultTheme != $_POST['X_DefaultTheme']) {// If not equal, update the default theme.
+		if($DefaultTheme != $_POST['X_DefaultTheme']) {			
 			// BEGIN: Update the config.php file:
-			$FileHandle = fopen('config.php', 'r');
-			if($FileHandle) {
-				$Content = fread($FileHandle, filesize('config.php'));
-				$Content = str_replace(' ;\n', ';\n', $Content);// Clean space before the end-of-php-line.
-				$Content = str_replace('\''.$DefaultTheme .'\';', '\''.$_POST['X_DefaultTheme'].'\';', $Content);
-				$FileHandle = fopen('config.php','w');
-				if(!fwrite($FileHandle,$Content)) {
-					prnMsg(_('Cannot write to the configuration file.'), 'error');
-				} else {
-					prnMsg(_('The configuration file was updated.'), 'info');
-				}
-				fclose($FileHandle);
-			} else {
-				prnMsg(_('Cannot open the configuration file.'), 'error');
+
+			$DefaultTheme = $_POST['X_DefaultTheme'];
+			// Path to the configuration file
+			$configFile = 'config.php';
+
+			// Read the contents of the configuration file
+			$fileContents = file_get_contents($configFile);
+			if ($fileContents === false) {
+				die("Failed to read the configuration file.");
 			}
+			
+			// Use a regex to find and replace the $DefaultTheme line
+			$pattern = '/^\s*\$DefaultTheme\s*=\s*["\'].*?["\'];/m';
+			$replacement = "\$DefaultTheme = '$DefaultTheme';";
+			
+			if (preg_match($pattern, $fileContents)) {
+				// Replace the line
+				$updatedContents = preg_replace($pattern, $replacement, $fileContents);				
+			} else {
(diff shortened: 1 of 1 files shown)


dalers:
Hi @bcinteractive. I'm jumping in here. Is there an issue for this PR? I recall a number of conversations over the year end (Discuss or issue topic or a comment to PR), but I didn't follow closely. I also don't profess to understand theming or "my theme" vs "global" theme (in config.php).

Can you please give more context so we can all learn. What behavior did you expect should update config.php and isn't, and how does this PR resolve the issue? Please sensei... ;-) :-)

dalers:
@bcinteractive  I tested commit 99fe7232c119255cc275fd1c2891350c06f3c098 (by @pakricard 2025-01-05 "Sanitize data") and config.php is updated when I change the default theme in [Main Menu > Setup > General Setup Options > System Parameters]. 

The page isn't updated with the new theme name after updating, but I viewed config.php after changing the setting (and back again), and if you navigate to somewhere else in the menu and then back to System Parameters, the new (correct) theme name was shown.

Perhaps the cause is unique to your system, and the commit fixes it on your system, but wanted to let you know it doesn't seem necessary for me (unless of course, the same fix has been committed seperately from this PR).

Cheers,
Dale

bcinteractive:
@dalers Thanks for testing.  It may have something to do with my config.php file -- as far as I know, my config.php file came from v4.15.2 - but ok.  Please disregard if this is not necessary.

Also, my title is misleading so I modified it to mean in System Parameters.

bcinteractive:
@dalers Please help me  double check. :smile:   I'm using the current master - while it's true that the cofig.php file is updated with the new selected theme every time the **update** button is clicked, the default theme input box doesn't reflect the change.

bcinteractive:
Hi @dalers,

Just an update, I noticed that while the file config.php is being updated with the new theme, the display in the Systems Parameters does not reflect the changes.  It does so when you make a refresh or return back to system parameters page after moving out of it.

Ben

pakricard:
Hi @dalers: 
I can't test this config.php issue as my config.php has been modified and no longer use $DefaultTheme.
A way forward for this could be if you could test this PR on your system. If your system works the same (or better) and at the same time, it fixes @bcinteractive issue, I think we can merge the PR.

dalers:
> ... if you could test this PR on your system.

IIuc, I will need to create a remote on my server for your GitHub fork, then pull the specific commit in the PR.

Does that seem correct?

pakricard:
Hi @dalers : 
I can't help much here... This PR is not on my branch, I guess it's probably easier just download the modified file and test, but really, I am no expert at all on the weberp installation process or github.

dalers:
> This PR is not on my branch,

Yes, of course you are correct. I will have to pull commit from @bcinteractive fork.

bcinteractive:
> Hi @dalers : I can't help much here... This PR is not on my branch, I guess it's probably easier just download the modified file and test, but really, I am no expert at all on the weberp installation process or github.

Hi @pakricard : the issue has nothing to do with installation process or github.  You can verify the behavior with your current copy of webERP.  Here's how you can recreate (or not) the behavior

1. Just log into your copy of webERP as admin
2. Go to Settings/Setup --> System Parameters
3. Zero in on the option for Default Themes (it's the number 3 item on the page)
4. Choose a different default theme
5. Click update  (this should result in your chosen default theme)

If you can NOT recreate the behavior then you can disregard this PR.

Thanks for looking into this.

dalers:
I tested with the current head of master (commit 4dacc8c1cf006a9d8cca076c8c56e1a00b4bf939) and found this issue still does not occur (i.e. config.php is updated when the default theme is changed using the System Parameters page - same as my earlier report https://github.com/timschofield/webERP/pull/345#issuecomment-2574436628).

pakricard:
Hi @bcinteractive :
I can't test this config.php issue as my config.php has been customized and no longer use Default Theme. So I can't recreate it, but it doesn't mean that the PR is not valid.  You might have the solution for a general issue here ;-).
That's why I was asking @dalers or anyone else who might be running a standard config.php file to check it out. Once we have confirmation, then we can commit it.

timschofield:
The $DefaultTheme variable isn't actually used anywhere, so unless anyone knows differently it should be removed entirely.

Tim

dalers:
Glad to hear from you again Tim!

On Jan 30, 2025 11:31 AM, Tim Schofield ***@***.***> wrote:
The $DefaultTheme variable isn't actually used anywhere, so unless anyone knows differently it should be removed entirely.
Tim

timschofield:
The $DefaultTheme variable isn't actually used anywhere, so unless anyone knows differently it should be removed entirely.

Tim

dalers:
I'm going to close this PR presuming it is now stale. @bcinteractive please comment if you feel this is not the case, and create a new PR.

Also see https://github.com/timschofield/webERP/discussions/461

PR #360, opened 2025-02-13
Demo mode warning msg in WWW_Users: Fix typo and improve language

andrewcouling:


timschofield:
Thanks, this message wasn't being shown because footer.php wasn't being included. I added in the include, but never even looked at the error message!

PR #397, opened 2025-03-28
Fix: include [footer|header].php instead of *.inc

aingelc12ell:
Revise include('includes/footer.inc') and include('includes/header.inc')


timschofield:
I have merged this. Thanks very much

Tim

PR #412, opened 2025-05-06
Update INSTALL and UPGRADING docs

dalers:


dalers:
Hi @timschofield hopefully these updates don't have major errors or omissions. There was some duplication before but now INSTALL.txt and UPGRADING.txt are essentially distinct. Fyi, I concentrated on installing and upgrading using the current database update system and web installer, and less on the older content.

There is still duplication between doc/INSTALL.txt and ./README.md, but it's minor and appears to be consistent. Also the ./LICENSE file might as well be moved to doc/ to be together with LICENSE.txt, but I didn't notice it until after creating the PR.

Cheers,
Dale

timschofield:
Thanks very much for this Dale

Issue #415, opened 2025-05-13
"Select a Different BOM" link in BOMs.php does not work


dalers:
The link "Select a Different BOM" in BOMs.php is not clickable.

![Image](https://github.com/user-attachments/assets/e22b0a4c-2b39-43d7-b86a-ba40ca1271c0)

dalers:
Resolved in PR https://github.com/timschofield/webERP/pull/416

dalers:
Tested and confirmed resolved.

PR #416, opened 2025-05-13
"back-link" corrections

dalers:
Resolve issue #415 (and for a bonus, correctly style the "Back to Items" link in Prices.php)


timschofield:
Thanks @dalers

PR #435, opened 2025-06-20
set correct requirements in composer.json (min php 8.1); refresh vendors

gggeek:
Since it was not very clear from the docs what is the minimum required version of php, I have:
- examined the min php versions from dependencies in composer.json
- figured out it is php 8.1
- improved composer.json to make that explicit + make sure it never downloads anything not compatible with it
- run a composer reinstall to "downgrade" dependencies to a set which works with php 8.1
- updated the readme

If this is not correct, and php 8.2 is the target minimum version, I can send a different, equivalent PR


gggeek:
Side notes:
- I set a big list of php extensions as recommended, but some of those might be declared as required in fact
- Is it correct that this can run on both mysql/mysqli/pgsql php exts, and both mysql/mariadb/pgsql dbs? If it is, we could update the docs; if not, there is some code which could be removed.

dalers:
Re min PHP version, we brag a bit about compatibility back to PHP 5.1 (per project readme), but I'm good with 8.1 if that is the common minimum version for the dependencies (all installed btw using Composer ;-) ).

dalers:
> * I set a big list of php extensions as recommended, but some of those might be declared as required in fact

I am not surprised.

> * Is it correct that this can run on both mysql/mysqli/pgsql php exts, and both mysql/mariadb/pgsql dbs?

Yes, although personally I'm unclear if mysql is still needed. I support less code is better and in a php 8.x world, is supporting mysql still relevant?

pakricard:
Last webERP version being PHP5 was 4.15.2 (dec 2021). Since some months ago, all the code in GitHub requires PHP8, and we named it webERP 5.0.0.rc. So, you are correct, now webERP requres PHP8.1

PR #446, opened 2025-06-22
Gg/issue 406 error in api manual

gggeek:
This should fix both the manual page for the api (in 3 languages), as well as the api itself.

There are two "todo" lines in the code - I might get around implementing that tomorrow.

What this doe not do is: abandon the phpxmlrpc api v3 in favour of using everywhere the v4 - that will have to be done in a further improvement pass.

ps: if this looks big is because the pxr lib has been moved to being managed by Composer, inside the vendors folder


gggeek:
Abandoning the phpxmlrpc api v3 in favour of using everywhere the v4: in progress...

PR #447, opened 2025-06-22
fix links and broken markdown; format for better readability in plain…

gggeek:
…text; fix spelling; 2024 -> 2025


dalers:
thank you @gggeek

PR #448, opened 2025-06-22
fix the badges in the readme, broken in the previous commit

gggeek:
Sorry for that


dalers:
Thanks for catching this @gggeek .

PR #454, opened 2025-06-23
Improve the html of multilingual manuals

gggeek:
- add charset declaration to the german one
- remove spurious html closing tags from the chinese ones


gggeek:
Hold on while I test this one more...

Issue #440, opened 2025-06-20
move to tcpdf 6.10.0, include it via composer


gggeek:
It seems that the currently included version of tcpdf has known security issues

See https://packagist.org/packages/tecnickcom/tcpdf

dalers:
Something seems wrong. Iiuc, tcpdf was fully replaced with DomPDF.

Wiki page: https://github.com/timschofield/webERP/wiki/TCPDF

Final(?) DomPDF commit: a043bb4af5360caa8c6e30d95b10a1433ae23b5b

dalers:
Also see [How to "USE DomPDF to create PDF reports rather than TCPDF"?](https://github.com/timschofield/webERP/discussions/405)

gggeek:
I ll double check then - if no tcpdf is used I'll drop it from the 'includes' folder

gggeek:
Sadly, there are still usages of TCPDF here and there, through classes which descend from TCPDF.
Even more worryingly, there is at least one usage of the previously-used library FPDF!

Not sure what is the best course of action here:
- first mode tcpdf to be a composer dependency, drop it later, or
- skip that and work instead on finishing the removal of TCPDF

gggeek:
Btw, looking at tcpdf changelog, it seems that there are no Breaking Changes reported between 6.7.5 and 7.10.0 - see https://github.com/tecnickcom/TCPDF/blob/main/CHANGELOG.TXT

pakricard:
> Sadly, there are still usages of TCPDF here and there, through classes which descend from TCPDF.
> Even more worryingly, there is at least one usage of the previously-used library FPDF!
> 
> Not sure what is the best course of action here:
> - first mode tcpdf to be a composer dependency, drop it later, or
> - skip that and work instead on finishing the removal of TCPDF

I would prefer the first option,  but not dropping it later on. 
Why?  
As webERP has used TCPDF for many years, there is a strong chance that some users have created their own reports over it (I am one of them, and probably not the only one). Dropping it would force us to rewrite these custom scripts.
If there is no security risk, why not leave it?
R

gggeek:
@pakricard I see your point. I'll start with moving it, and keeping it locked at the current version. Then, when we feel safe, we can update it easily to the latest version

gggeek:
[...]

PR #455, opened 2025-06-23
Move tcpdf to composer (keep same version as it was)

gggeek:


gggeek:
First step on the road to fully implement #440

PR #468, opened 2025-06-25
improve comments and formatting of config.distrib.php

gggeek:


dalers:
Why was ".htaccess" renamed ".htaccess_"?

timschofield:
> Why was ".htaccess" renamed ".htaccess_"?

I don't really understand why we need a .htaccess file in the project. It isn't used by nginx servers anyway, but I feel this is something for the sysadmin to be doing, and not something webERP should be mandating.

dalers:
> It isn't used by nginx servers anyway,

Is webERP served _only_ by Nginx? Iiuc Nginx and Apache market share is roughly equal (fwiw, I'm using Apache for both PHP and Python web apps to have a common hosting app with less to learn and stay current with).

> I feel this is something for the sysadmin to be doing, and not something webERP should be mandating.

the webERP project is not mandating anything, a sysadmin can alway modify the .htaccess if they want. Imho, the goal of a new webERP sysadmin is to get webERP running, securely, for business users, and in my experience, the less they have to learn and do, the happier they are and fewer mistakes get made.

If the webERP project can provide assistance in the way of a .htaccess file so the sysadmin doesn't have to craft one themselves (after first figuring out how webERP expects itself to be served), isn't that a good thing? Doesn't it help webERP to be adopted and used more readily?

I would counter by arguing if Nginx is a common server for webERP, then we should be adding Nginx config not removing Apache config.

dalers:
I should clarify, I only want to know why .htaccess was renamed before I would be willing to merge the PR. 

@gggeek , Iiuc, by renaming .htaccess Apache will now be able to access files in includes/. Is this what was intended? I thought the goal was for Apache to _not_ be able to access files in includes/ (i.e. that it should be possible to access files in includes/ _only_ by being included from within a file in the root directory). What am I missing?

timschofield:
I believe the job of webERP is to supply the webERP code, not to get involved in server configurations. Of course we can supply docs with examples of how to setup a server, but in the end, server configuration files don't belong in the webERP code.

gggeek:
> @gggeek , Iiuc, by renaming .htaccess Apache will now be able to access files in includes/. Is this what was intended? I thought the goal was for Apache to not be able to access files in includes/ (i.e. that it should be possible to access files in includes/ only by being included from within a file in the root directory). What am I missing?

You are not missing anything. I did miss that silly renaming before committing and pushing. The power of many eyes in action :-D

gggeek:
> I believe the job of webERP is to supply the webERP code, not to get involved in server configurations. Of course we can supply docs with examples of how to setup a server, but in the end, server configuration files don't belong in the webERP code.

Sorry, but I disagree with this. 

No-one reads the docs carefully. It's a fact of life.

If they are terse, people will fall through the cracks. If they ar detailed, people skim over them and still get stuff wrong.

Docs are hard to keep up to date and always in sync with the code, so they are often behind the latest developments. That's why documenting the code via phpdoc, markdown or similar tooling has spread like wildfire - keeping the docs close to the source code is a good thing.

.htaccess files can be considered the "as close as possible to the source" documentation, which is easy to maintain by the developers, and which any half competent web server admin can use to infer which directories we, the developers, consider to be safe for access from the web and which ones should not be.

They are harmless when not using Apache, or when Apache has disabled support for them.

They do add some security when in effect, following the defense-in-depth principle.

They do not introduce a security risk, as long as we do not _only_ rely on their presence to keep files secret. Which is not my intention to do.

On a similar vein, in the php code, there are lines that call "ini_set". Would you consider that this is also crossing the "configuration line", and that php.ini should be only the concern of the sysadmin? Vice-versa, it is quite customary to use `config.php` to define platform-specific concerns, such as f.e. the setup of the logging system, server timezone, directories where to store session files, etc. What about those?
The boundary between webserver and app-server, config and code, hosting and application is porous.

gggeek:
btw, rename of .htacces reverted in this branch

timschofield:
No. We produce ERP software. Server configuration is out of our remit. By pulling this file into the main branch whilst discussions were still ongoing we now have the situation where a sysadmin who has created their own .htaccess file will have it overwritten without any warning by something that we as a project can't even agree upon amongst ourselves. 
We produce ERP software, the sysadmins run their servers.
I am happy to produce example htaccess files in the docs to act as a guide, but if people want to ignore them, then that is their prerogative.
I really feel that a PR shouldn't be pulled while discussions are ongoing, or the whole code base will end up in an anarchic mess.

Tim

gggeek:
> I really feel that a PR shouldn't be pulled while discussions are ongoing, or the whole code base will end up in an anarchic mess.

Sorry, I do not think anyone intended this. The fact is: the `.htaccess` files went in in prior PRs, which apparently flew under your radar. This PR, which @dalers merged, only changed the `config.distrib.php` file.

I am happy to send in another PR to remove all of the added `.htaccess` files, if you feel so strongly about those.

> a sysadmin who has created their own .htaccess file will have it overwritten without any warning

I think this pattern is worth a more in-depth discussion: how much, and how, are end-users supposed to customise the webERP installation, and how can those changes survive an upgrade?

If they hacked any of the .php / js / css / html / sql files we provide, then their changes will be overwritten, when they copy the new source over them.

If they added their own .php / js / css / html / sql, there is a chance that we also might add files with the same name, and also end up overwriting them.

This is a common problem, and the generally accepted solution is that the whole tree of source code, with all its subfolders, is under the sole control of the application. Adding any random file, including `.htaccess` ones, is not supported. Means have to be provided for end users to add in their changes in the form of "plugins", which reside in specific folders, the same way the installation-specific configuration files reside in their own place. This is also btw the reason why I proposed, elsewhere, to move both the `config.php` and `companies` folder to be moved within a single `config` top-level folder: to make it easier to separate end-user stuff from webERP stuff.

gggeek:
@timschofield revert of .htaccess sent in PR #483

PR #472, opened 2025-06-25
remove mariadb from list of supported db "types"

gggeek:
Atm there is a bit of confusion about what the `$DBType` configuration is supposed to stand for:
- we accept postgresql, mysql, mysqli, mariadb as valid values
- but mysql and mysqli are in fact php-side connectors, and they work both with mysql and mariadb,
- while mariadb is a server-side component, and the webERP php functions used to connect to it are the same as the `mysqli` ones - except for the last 2 commits which I made in the last couple of days to `mysqli` only

I think this confusion has to be cleared up, esp, if we later want to introduce a pdo-based connector - we do not want to end up with the combinatorial explosion of mysql, mysql-mysqli, mysql-pdo, mariadb-mysql, mariadb-mysqli and mariadb-pdo. It also makes little sense to duplicate all of the `mysqli` connector code within a `mariadb` file.

If there are differences between mysql and mariadb behaviour which will need php code to be adjusted, I think that instead of providing 2 drivers we'd better have 1 driver which can check the server type dynamically and take the appropriate decision.

There are also newfangled databases out in the wild today which are (supposedly) mysql compatible, I do not think that we want to add a new complete connector for each - that would be a lot of duplication.

So the result is: drop the current "mariadb" connector, update the docs reflecting the fact that mariadb rdbms is supported, and use either the mysql or mysqli connector for it


timschofield:
MariaDB and mysql are diverging. For instance the syntax for calculated fields is very different in both. This means that we need different UpgradeDB_mariadb.php and UpgradeDB_mysqli.php files so we need a different $DBType. You are right that at the moment we don't need separate ConnectDB_mariadb.php and ConnectDB_mysqli.php, but I would prefer to keep both types.
I would agree that the postgres files can go, they don't work anyway.

Thanks
Tim

gggeek:
> MariaDB and mysql are diverging. For instance the syntax for calculated fields is very different in both. This means that we need different UpgradeDB_mariadb.php and UpgradeDB_mysqli.php files so we need a different $DBType

You are right. I noticed at the moment I was deleting the `UpgradeDB_mariadb.php` file that it had a 1 line difference from `UpgradeDB_mysqli.php`.

I have thought about how to best tackle the problem. I think I see the following possible solution:
* keep "dbtype" as a separate variable from "dbdriver"
* for "dbtype", I would recommend letting the php code figure it out via queries. But it could be done via a setting in config.php + a setup wizard drop-drown
* for the db-related code, I'd create classes which group the existing functions into static methods
    * one class which holds methods to connect, execute queries, manage transactions
    * one class which holds methods to manage the data dictionary, ie. create tables/indexes and co
* at the moment, the only difference between mysql and mariadb is in the latter class, in the function to create virtual columns. As this code is not loaded or executed on every page view, I think that we can afford to have the data-dictionary-manager run an extra query to figure out the db type
* presuming the differences between mysql and mariadb will evolve over time, I would take the approach of using subclassing, to keep maximum flexibility while reducing code duplication. Ie. the mariadb schema manager class would be a descendent of the mysql one
* the code of those classes will not inherently start any connection when included. This means that it could be reused both by the standard controllers and by the setup wizard one which creates the db, with the added bonus of making it possible to use mysql (instead of mysqli), and possibly in the future pdo_mysql, at installation time  

WDYT?

dalers:
> Atm there is a bit of confusion about what the `$DBType` configuration is supposed to stand for:
> 
> * we accept postgresql, mysql, mysqli, mariadb as valid values
> * but mysql and mysqli are in fact php-side connectors, and they work both with mysql and mariadb,
> * while mariadb is a server-side component, and the webERP php functions used to connect to it are the same as the `mysqli` ones - except for the last 2 commits which I made in the last couple of days to `mysqli` only

I agree there is confusion, resulting from lack of orthogonality in the terms and their context (imho not uncommon when software evolves organically) and cleaning this up will certainly pay dividends in the future if the DBMS is further abstracted, although I am still not convinced webERP _needs_ to support more databases and e.g. ONLY supporting MariaDb (hypothetically) would also simplify the problem.

However, your proposed classes and subclasses is unfortunately not something I can comment on from experience. :-(  I learned programming in the 80's with FORTRAN, COBOL and C, and what I know about OO comes from a couple Arduino projects. I'm working on a couple PHP courses in LinkedIn Learning (our local library card gives us access), but I fear I don't have enough years left to become an even half-decent web dev!

timschofield:
This feels to me like a solution in search of a problem. The current system is simple and works. Getting the code to work out what DB is being used just seems over complicated, and doesn't address any problem that currently exists.

Tim

gggeek:
@timschofield your call.

I think the current database subsystem has multiple problems.

Off the top of my head:
* the confusion between "dbtype" and "dbdriver", 
* the lack of support for pdo, 
* the fact that, by not querying the DB for its kind we'd force the end user to specify more configuration than is necessary and to change it when upgrading the database, 
* the absent usage of prepared statements in SQL queries, 
* the fact that it delegates prevention of SQL injection to the `session.php` file, thus making SQL injection possible as soon as php code is run via the command line or via a non-http-form input, 
* the fact that the DB classes can not be used as is by the installer code because they conflate functionality declaration with connection initiation,
* the fact that they have code in them which ends execution and displays html output, making them unsuitable for usage within a pdf/csv/xmlrpc context,
* they presume that $_SESSION is always set,
* they do not support nested transactions,
* and the fact that we could wholesale replace 90% of their code with an external library, thus reducing maintenance burden while gaining more features.

That's before even starting to discuss how we could provide support for sqlite/postgres by moving away from plain sql files for database buildup and move to a db-agnostic solution.

Fixing those issues would imho justify the cost of refactoring and the added code complexity. 

PS: I might be misunderstanding your comment here - I am not sure if, out of my proposal, you think that the auto-db-type-check is overkill and the rest is fine, or that it is just all unnecessary complication.

gggeek:
Since this PR might never be merged, I created PR #484 as a non-controversial improvement in the meantime

timschofield:
My previous remarks had been aimed at the auto-db-type-check which seems overly complicated. If the user says they are using MySQL or MariaDB I think we should just take their word for it. If they are wrong, then that is there problem and they should rectify it. If somebody changes DBMS on an already established system then I would assume they know what they are doing, and a small configuration change in config.php is the least of their problems :)

However.....

> Off the top of my head:
> 
>     * the confusion between "dbtype" and "dbdriver",

I'm not sure I really see this confusion. Perhaps we should say MySQL instead of MySQLi in the installer to make it clearer?

> 
>     * the lack of support for pdo,

Having a ConnectDB_pdo.php file which exposed the same functions as the current mariadb and mysql files would be interesting - if it is technically feasible?? One major problem we have is that despite the best intentions a lot of mysql specific SQL has crept in over the years. I had a look at this problem once, and the solution is not trivial. There are currently 4262 SQL queries currently in our code so examining them all is a big task.

> 
>     * the fact that, by not querying the DB for its kind we'd force the end user to specify more configuration than is necessary and to change it when upgrading the database,

Well the user has to select it when installing, and after that there is no more configuration unless they choose to change DBMS in which case as I said above a one line change in config.php is the least of their problems.

> 
>     * the absent usage of prepared statements in SQL queries,

This has been discussed ad nauseum in the past, and been rejected. There are a huge number of queries in webERP which would have to be rewritten, to end up with less readable code for no real gain.

> 
>     * the fact that it delegates prevention of SQL injection to the `session.php` file, thus making SQL injection possible as soon as php code is run via the command line or via a non-http-form input,

Can you give an example of when this would happen?

> 
>     * the fact that the DB classes can not be used as is by the installer code because they conflate functionality declaration with connection initiation,

Mostly the installer does use them. It's largely the connection part that was done separately.

> 
>     * the fact that they have code in them which ends execution and displays html output, making them unsuitable for usage within a pdf/csv/xmlrpc context,

The thinking here was that if there is an error, the code should just stop and post an error to the screen, rather than continue to output to the PDF or CSV file

> 
>     * they presume that $_SESSION is always set,

Isn't it always?

> 
>     * they do not support nested transactions,

Which we don't use, though if we wanted to they could be changed to.

> 
>     * and the fact that we could wholesale replace 90% of their code with an external library, thus reducing maintenance burden while gaining more features.

The maintenance burden on the ConnectDB_XXXX.php files is not onerous. The last time a new function was added in was Feb 2018, and changes since then have not been to the functionality.

> 
> 
> That's before even starting to discuss how we could provide support for sqlite/postgres by moving away from plain sql files for database buildup and move to a db-agnostic solution.

Supporting other databases was always an aspiration, but I think both Phil and I came to the conclusion it wasn't going to happen. Actually creating a ConnectDB_sqlite.php file wouldn't be that tough, the hard bit would be changing all the queries with MySQL specific SQL in them.

> 
> Fixing those issues would imho justify the cost of refactoring and the added code complexity.
> 
> PS: I might be misunderstanding your comment here - I am not sure if, out of my proposal, you think that the auto-db-type-check is overkill and the rest is fine, or that it is just all unnecessary complication.

gggeek:
@timschofield 

> I'm not sure I really see this confusion. Perhaps we should say MySQL instead of MySQLi in the installer to make it clearer?

One can access a mariadb with either the mysql, mysqli and pdo php extensions.
Also, a mysql server, or any mysql-compatible server such as offered by AMZ, can also be accessed with either the mysql, mysqli and pdo php extensions.

The current code does mix the two concepts.

If we are to improve the situation for the future, so that it is possible to add support for further db-drivers, and different db-servers I think that we should split the two concepts.

I'm not entirely happy with the idea of having the user specify 2 separate fields in the setup wizard, but that is not the end of the world - in the grand scheme of things it would still be a step forward.

> Having a ConnectDB_pdo.php file which exposed the same functions as the current mariadb and mysql files would be interesting - if it is technically feasible?? One major problem we have is that despite the best intentions a lot of mysql specific SQL has crept in over the years. I had a look at this problem once, and the solution is not trivial. There are currently 4262 SQL queries currently in our code so examining them all is a big task.

It is entirely possible to implement the current DB API on top of PDO. Just not use prepared statements ever - you loose one of the biggest benefits of PDO, but you allow more flexibility in the setup (not sure there are any setups out there with ext_mysqli disabled and ext_pdo_mysql enabled though. That, and claiming we support pdo might let users falsely assume we use prepared statements, too)

> Can you give an example of when this would happen?

As I already did: in the xmlrpc api

> Mostly the installer does use them. It's largely the connection part that was done separately

Still, this is atm the only reason that, atm, webERP v5 will require the mysqli extension and not work with the mysql extension (side note: I'd be happy to drop support for the mysql extension, but you disagreed with that in another post), and could be fixed by improving the DB classes

> The thinking here was that if there is an error, the code should just stop and post an error to the screen, rather than continue to output to the PDF or CSV file

The fact is, there might be output buffering at play, so the message would not be shown to the end user in any case. Or, the end user might be an API consumer, not a browser. Or, we could want to implement a cli script, in which case the output should not be formatted as HTML. What I think is a better approach to the current code is to have the DB classes throw an exception if they fail to connect, but let the upper layers in the code (maybe by installing an error handler, which is btw a very good idea in general) decide how to display such error.

> Isn't it always?

It might be at the moment, but that prevents the development of cli scripts, apis which are stateless and need no session, and a slew of other things. Of course, the DB classes are not the only ones which would need to be fixed in order to make that possible, but it could be a 1st step in the good direction. Also, making code isolated from the the current state, which is what `$session` is, is a core good development principle.

timschofield:
> One can access a mariadb with either the mysql, mysqli and pdo php extensions. Also, a mysql server, or any mysql-compatible server such as offered by AMZ, can also be accessed with either the mysql, mysqli and pdo php extensions.
> 
> The current code does mix the two concepts.
> 
> If we are to improve the situation for the future, so that it is possible to add support for further db-drivers, and different db-servers I think that we should split the two concepts.
> 
> I'm not entirely happy with the idea of having the user specify 2 separate fields in the setup wizard, but that is not the end of the world - in the grand scheme of things it would still be a step forward.

I'm still not sure I understand the problem here. When installing webERP the user picks whether they are using MySQL or MariaDB, and then everything works from there. I think two fields will only confuse the user for no obvious gain.

> 
> It is entirely possible to implement the current DB API on top of PDO. Just not use prepared statements ever - you loose one of the biggest benefits of PDO, but you allow more flexibility in the setup (not sure there are any setups out there with ext_mysqli disabled and ext_pdo_mysql enabled though. That, and claiming we support pdo might let users falsely assume we use prepared statements, too)

Using prepared statements has been discussed in the project several times before and the idea was discarded.

> 
> As I already did: in the xmlrpc api

True, I tend to forget about the API. I would suggest fixing the API rather than opening the can of worms of changing session.php

> Still, this is atm the only reason that, atm, webERP v5 will require the mysqli extension and not work with the mysql extension (side note: I'd be happy to drop support for the mysql extension, but you disagreed with that in another post), and could be fixed by improving the DB classes

I think I disagreed with removing ConnectDB_mysql.php because some people may still wish to use it. I would advise against it, but this is free software and people should have the freedom to use it as they want.

gggeek:
> I'm still not sure I understand the problem here. When installing webERP the user picks whether they are using MySQL or MariaDB, and then everything works from there. I think two fields will only confuse the user for no obvious gain.

I agree that 2 fields might be confusing to the user.

Still, when you say "this is free software and people should have the freedom to use it as they want", and when there is an open discussion ticket about "adding support for the pdo driver", it makes me think that you would agree that providing 3 drivers (based on the mysqli, mysql and pdo php extensions) is a good thing - it gives users more freedom, it allows to install on a more wide set of servers, and it future-proofs the software, in case any of those extensions gets deprecated or abandoned.

Unrelated to the choice of driver to use, there are _many_ databases which expose a "mysql-like" API, ie. they speak the mysql wire protocol, but generally have a different feature set and support slightly different dialects of sql. An example post listing some of those: https://medium.com/@shenli3514/the-landscape-of-mysql-wire-compatible-database-c7d1838f79ef

Do we want to give users complete freedom of choice in that regard? If we stick to the same principle, we should.

The way I mentally model this is that we have a matrix of M ("drivers") x N ("database types") combinations possible. As it clearly would be pointless to create so many php scripts/classes to handle all permutations, I would go for creating one "DB" class for each "driver", and possibly subclass it, or use "switch" statements here and there for the handling of the differences between servers.

The current situation on the other hand is the following:

* only the mysqli extension is fully supported, ie. we have one "driver"
* the mysql extension is partially supported, but not by the installer, so it might be as well dropped
* it seems not possible to use the mysql php extension to connect to a mariadb server, even though that is not a php or database limitation 
* the "mariadb" DB file duplicates 99% of the code of the mysqli one

So, yes, "it works". But also, "it could be handled better"

gggeek:
> I think I disagreed with removing ConnectDB_mysql.php because some people may still wish to use it. I would advise against it, but this is free software and people should have the freedom to use it as they want.

If I get your idea correctly, it is: leave in the codebase a DB connector which can not be picked at install time, which we most likely will not test using automation, which can only be used to talk to mysql servers and not mariadb ones (as it does not have proper handling of DDL) and does not even clearly state that on the tin?

On a side note: I will provide a set of code branches where I show how it is possible to achieve a 100% backwards compatible refactoring.

That will not necessarily be to anyone's liking, as there will be more php files that there are atm, rather than less, and it might be perceived as unnecessary complication.

But it will provide a way for us to start adding deprecation warnings right now, and be able to act on those deprecations (ie. remove the compatibility shims) later on, in the distant future.

timschofield:
> > I think I disagreed with removing ConnectDB_mysql.php because some people may still wish to use it. I would advise against it, but this is free software and people should have the freedom to use it as they want.
> 
> If I get your idea correctly, it is: leave in the codebase a DB connector which can not be picked at install time, which we most likely will not test using automation, which can only be used to talk to mysql servers and not mariadb ones (as it does not have proper handling of DDL) and does not even clearly state that on the tin?
> That will not necessarily be to anyone's liking, as there will be more php files that there are atm, rather than less, and it might be perceived as unnecessary complication.
> 
> But it will provide a way for us to start adding deprecation warnings right now, and be able to act on those deprecations (ie. remove the compatibility shims) later on, in the distant future.

I'm still a bit bemused by this. As I see it we currently have a ConnectDB_mysql.php that some current users may or may not still be using and we have ConnectDB_mysqli.php and the ConnectDB_mariadb.php files that the user will use depending on their selection of DB type in the installer. Removing the mysql file risks possibly breaking some existing installations while not giving any benefits to the rest of the users.

> 
> On a side note: I will provide a set of code branches where I show how it is possible to achieve a 100% backwards compatible refactoring.
> 

I'm sure you can do this but I don't understand why you would.

Tim

gggeek:
> I'm still a bit bemused by this. As I see it we currently have a ConnectDB_mysql.php that some current users may or may not still be using and we have ConnectDB_mysqli.php and the ConnectDB_mariadb.php files that the user will use depending on their selection of DB type in the installer. Removing the mysql file risks possibly breaking some existing installations while not giving any benefits to the rest of the users.

I tried my best to explain why I think that the current situation is not optimal, and lay out possible paths for improvements, with overly lengthy posts going in great technical detail.

You seem neither to be convinced that the current situation is sub-optimal, nor that it could be improved.

I will now close this PR, as I don't want to waste anyone's time. The discussion will remain available for anyone willing to revisit this topic in the future, via a search on github.

gggeek:
> I'm sure you can do this but I don't understand why you would.

Thanks for this eye-opening answer.

Indeed, if there is no chance that refactoring will be accepted, pushing it would only be satiating hubris.

timschofield:
> > I'm sure you can do this but I don't understand why you would.
> 
> Thanks for this eye-opening answer.
> 
> Indeed, if there is no chance that refactoring will be accepted, pushing it would only be satiating hubris.

I apologise if my remarks caused offence, that wasn'y my intention. I was trying to make a robust defence of code that has worked without problem for many years. Supporting other databases was always an aspiration, but as I have said before the major problem with this is that a lot of MySQL specific SQL has crept in over the years. Fixing this would be the start for any move to supporting other DBMS.

gggeek:
> I apologise if my remarks caused offence, that wasn't my intention

No offence taken.

> Supporting other databases was always an aspiration, but as I have said before the major problem with this is that a lot of MySQL specific SQL has crept in over the years. Fixing this would be the start for any move to supporting other DBMS.

I agree with this. I think I even mentioned, in the GH discussion about PDO, that trying to make an application truly database independent is a huge feat, and that it might not even be a goal worthy of pursuit, ever, for webERP, esp. given the current state of things (code + dev team + project history).

Otoh, what this PR, and its accompanying discussion, are about, is a different thing. It is: "keep supporting only mysql, mariadb and the other mysql-compatible RDBMS, but allow the app to use either of the mysql, mysqli and pdo php extensions for connecting to the database".

Thinking more about your statement of "do not ask the users too many choices to avoid confusing them", I think that another solution to this conundrum, which might possibly appeal to everybody is:

* only let the users pick the RDBMS type at install time
* in the config.php file , have 2 separate settings, one for the php extension to use, and one for the type of rdbms
* let the installer pick the php extension to use in an automated way, ie. by order of preference (while checking their presence): 1st mysqli, 2nd pdo, 3rd mysql

The "php extension to use" variable would then be used by the code to decide which php files to include / which php classes to load, whereas the "type of database" variable would be used by that code to implement specific handling of DDL.

Needless to say, I'd still argue that, if we go down this route, this would also be the perfect time to refactor the existing DB related scripts...

PS: yes, I am aware that, as with almost any other refactoring, this will come as a surprise to existing users and that, if they had customised their installation in any way, they will have to spend some time figuring out how to port their changes to the new version of the ERP.

gggeek:
Note: the above comment was edited a few times after its original posting. I have a tendency to do that. I'm sorry that this might cause communication snags if anyone is reading the comment via an email notification and does not check its final version online.

(does anyone know how to work around that issue, apart from refraining from editing comments, once posted?)

PR #475, opened 2025-06-25
group together all installer sql files; …

gggeek:
…get license text from actual license file; move trigger sql outside of php code; use a bit more absolute paths


timschofield:
Merged, thanks.

PR #476, opened 2025-06-25
remove the BOM from the start of a php file

gggeek:
BOM is invisible but insidious ;-)


dalers:
Confirmed CountriesArray.php contains BOM

2025-06-26 13_59_25 CountriesArray php contains BOM

dalers:
File no longer has BOM after merge

2025-06-26 14_05_51 file no longer has BOM

PR #478, opened 2025-06-25
use function-call style calls to exit

gggeek:
Since php 8.4, `die` and `exit` have been transitioned from being language constructs to being proper function calls.

As such, it makes sense to always use parenthesis when calling them


timschofield:
This PR has significant conflicts. Probably due to my merging them in the wrong order. Probably best to redo it?

Tim

gggeek:
> This PR has significant conflicts. Probably due to my merging them in the wrong order. Probably best to redo it?

Indeed. Will do.

gggeek:
Closed in favour of #494

PR #484, opened 2025-06-26
port to the mariadb driver the fixes recently added to the mysqli one

gggeek:


timschofield:
Thanks. Pulled.

gggeek:
@timschofield it might be that it did nor merge because of a conflict. I have now fixed the conflict

PR #489, opened 2025-06-26
remove spurious semicolons

gggeek:


dalers (approved):
suspect spurious semicolons were unintended side effects of a global search and replace

pakricard (approved):

gggeek:
> suspect spurious semicolons were unintended side effects of a global search and replace

I'd put my money on copypaste ;-)

It is possible to check using `git blame` btw...

PR #501, opened 2025-07-20
clarify mail group name in error message

dalers:
Error message reported "There are no members of the Inventory Valuation Recipients email group" but the group name is actually "InventoryValuationRecipients" (no spaces).


timschofield:
Thanks Dale

PR #506, opened 2025-07-26
remove echoing of the utf8-bom

gggeek:
while at it, fix also:
- While -> while
- spaces to tabs
- echoing of the BOM for each line in csvs instead of only at beginning of file (now left at beginning but commented out)


timschofield:
Thanks Gaetano

PR #510, opened 2025-07-26
 add apple-touch-icon.png

gggeek:
Spec available at https://developer.apple.com/library/archive/documentation/AppleApplications/Reference/SafariWebContent/ConfiguringWebApplications/ConfiguringWebApplications.html#//apple_ref/doc/uid/TP40002051-CH3-SW4

Recommended image size: 180x180 px

I used the image from the website, resized and with transparent padding added


gggeek:
Note: we might as well add a link to it in the html, for when the site is not installed at the web server root...

PR #514, opened 2025-07-28
revamp the shell scripts in the build directory

gggeek:
Even if those are not used much atm, better to have them in working condition than broken and unsafe...

Also, splitting them in multiple parts allows their usage outside of a full build, eg. we could run php correctness checks automatically on PRs, and update of po files automatically before a new tag or release is done, etc...

### DIFF
--- a/build/TruncateAuditTrail.sql
+++ b/build/TruncateAuditTrail.sql
@@ -1,2 +1,2 @@
-USE weberpdemo;
+-- USE weberpdemo;
 TRUNCATE TABLE audittrail;
--- a/build/check_syntax.sh
+++ b/build/check_syntax.sh
@@ -1,17 +1,17 @@
-#!/bin/bash
+#!/usr/bin/env bash
 
-#ROOT_DIR=$PWD
-#cd $ROOT_DIR/..
-for f in `find . -name "*.php" -o -name "*.inc"`
-do
-    newname=`echo $f | cut -c3-`
-    filename="$newname"
-    echo $filename
-    output=$((php -l $filename ) 2>&1)
+BASE_DIR="$(dirname -- "$(dirname -- "$(realpath "${BASH_SOURCE[0]}")")")";
 
-    if [ $? != 0 ]
-    then
-		echo '**Error** '$output >> ~/weberp$(date +%Y%m%d).log
-		echo '' >> ~/weberp$(date +%Y%m%d).log
+cd "$BASE_DIR";
+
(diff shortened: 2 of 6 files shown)


timschofield:
I wrote they check_syntax.sh script many years ago when we were using SVN. These days I do syntax checking via a git pre-commit hook.

gggeek:
It might be good to add the commit hook code to the git repo, as long as it's not tied specifically to your computer's setup.

I have f.e. added this commit hook to another lib of mine: https://github.com/gggeek/phpxmlrpc/blob/master/.githooks/pre-push.sh

timschofield:
> It might be good to add the commit hook code to the git repo, as long as it's not tied specifically to your computer's setup.
> 
> I have f.e. added this commit hook to another lib of mine: [gggeek/phpxmlrpc@`master`/.githooks/pre-push.sh](https://github.com/gggeek/phpxmlrpc/blob/master/.githooks/pre-push.sh?rgh-link-date=2025-07-29T09%3A32%3A37Z)

Many years ago I did put the hooks I use here: https://github.com/timschofield/hooks

I tried to get some interest in using these as standard at the time, but there didn't seem to be much interest.

Tim

dalers:
> I tried to get some interest in using these as standard at the time...

@timschofield Do you mean before you took over the project? It would seem the hooks are "standard" now if I understand correctly.

timschofield:
Hi Dale, it was here:

https://web.archive.org/web/20230924213515/http://www.weberp.org/forum/showthread.php?tid=8348

With git, each developer can have theor own hooks, these were just mine.

Tim

dalers:
Thanks for the link @timschofield, it was a good overview of your hooks (in fact, you might consider copying the post into the README.me file in your hooks repo). 

I now understand hooks are configured locally and run client side (my mistake thinking they ran server side e.g. on a PR). 

1. Have you thought about moving the hooks into the project? It would go against the "only webERP code in repo" philosophy but would a) manage webERP project hooks as part of the webERP project, and b) be less work for developers to use.

2. Have you considered replacing the checks in the hooks with GitHub Actions? Iiuc, this would be the current best practices as it avoids depending on individual developers to configure their own systems (like me ;-))

timschofield:
>     2. Have you considered replacing the checks in the hooks with GitHub Actions? Iiuc, this would be the current best practices as it avoids depending on individual developers to configure their own systems (like me ;-))

To be honest I have never had the time o really look at what github actions do. I probably should

Tim

PR #533, opened 2025-07-31
fix link to create sales types in an error message; reformat miscfunctions.php

gggeek:


timschofield:
Thanks Gaetano.

Is it possible to have one issue per pull request?

I missed a problem in one PR yesterday because it was very big, and had multiple issues

gggeek:
> Is it possible to have one issue per pull request?

Yes, sorry. I'll try to stick to that

PR #535, opened 2025-07-31
add 2 missing images to arc-dark theme

gggeek:
the same icons seem to be used by multiple other themes, so I copied them over


timschofield:
I think it would be better to use next.png and previous.png instead of the svg files. They are themed so look better.

Tim

timschofield:
I am closing this PR as I have replaced the usage of the next.svg and previous.svg images from the system and replaced them with the standard png files

Issue #529, opened 2025-07-30
Bring back changelog and/or NEWS file


gggeek:
In view of the 5.0 release, some documentation of recent changes is welcome. Atm there seems to be nothing for the work done after 2022-05-15.

I suggest the following:

- do not produce any more a detailed changelog. Update file doc/CHANGELOG.md with a  note at the top stating that all recent changes are visible in github, with a link to https://github.com/timschofield/webERP/commits/master/
- add a NEWS file, which has a high-level description of the most important changes. Not automatically built from git commits, but curated by hand

I can start the list with the following:
- increased minimum php requirements to version 8.1
- deprecated support for php extension `mysql` to connect to the database (code for it is still there but it might be dropped in the future). Use `mysqli` instead
- db tables are now created using `utf8_mb4` character set if the database supports it (this is actually not yet merged ;-)
- rewritten the installer
- fixed the XMLRPC API
- updated pdf report writer TCPDF to version 6.XX
- introduced a new pdf report writer: DomPDF, to eventually replace all usage of TCPDF
- new and improved database upgrade system
- new login screen
- new agents commission system
- new dashboard system
- new system for processing regular payments
- new General Ledger budget system
- new budget system
- new popup context sensitive help system
- automatically log out the user, and then return them to the correct module when they log back in
- css styling changes
- various bugfixes

etc...

timschofield:
Hi Gaetano, I am more than happy for you to maintain either a changelog, or a news file. I know from experience I am rubbish at such things, and will always find something better to do than keep these it up to date :-) 

Tim

gggeek:
I think this cane be closed

timschofield:
I try and encourage the person who opened the issue to close it.

Tim

PR #558, opened 2025-08-06
add NEWS file, update README

gggeek:


gggeek:
Partial fix for #529 (the NEWS file has to be updated more, I will finish digging through the git log and send another PR in case)

PR #563, opened 2025-08-07
Remove all occurrences of $DbgMsg

gggeek:
As mentioned in comment https://github.com/timschofield/webERP/commit/9decfebdf9a62388f71117341e370287b19750ff#commitcomment-163569726

Side notes:

* this should reduce the size of the translation files quite a bit, once we get around to updating them
* this kind of refactoring exercise really makes me question the choice of not having a database-abstraction layer which would deduplicate the huge number of queries found in almost every php file

### DIFF
--- a/AccountGroups.php
+++ b/AccountGroups.php
@@ -17,13 +17,12 @@ function CheckForRecursiveGroup($ParentGroupName, $GroupName) {
 ie the parent group results in a recursive group structure otherwise false ie 0 */
 
 	$ErrMsg = _('An error occurred in retrieving the account groups of the parent account group during the check for recursion');
-	$DbgMsg = _('The SQL that was used to retrieve the account groups of the parent account group and that failed in the process was');
 	do {
 		$SQL = "SELECT parentgroupname
 				FROM accountgroups
 				WHERE groupname='" . $GroupName ."'";
 
-		$Result = DB_query($SQL,$ErrMsg,$DbgMsg);
+		$Result = DB_query($SQL, $ErrMsg);
 		$MyRow = DB_fetch_row($Result);
 		if($ParentGroupName == $MyRow[0]) {
 			return true;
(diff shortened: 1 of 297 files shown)


timschofield:
Having the SQL output to the browser via $DbgMsg when running in debug mode can frequently be very helpful when locating a bug. What is the advantage to the user in removing it?

Tim

gggeek:
What this PR does is not to remove the echoing of the SQL string, but to remove all the _translated labels_ which were used to accompany such echoing. eg: 

`The SQL that was used to retrieve the account groups of the parent account group and that failed in the process was`

will now simply be

`The SQL that failed was`

What we gain is a massive reduction in code size. I think that it outweights the loss in Developer eXPerience

timschofield:
Well I would say that the first error message is slightly more useful than the second, in that it narrows down where the error is happening.
I may be missing something but it seems to me that we have a "monster pull request" that makes the code slightly less useful for both users and developers, in order to slightly reduce the code size.
Most of these messages have been around for 20+ years with no previous complaints from any users that I recall.

Tim

gggeek:
> Well I would say that the first error message is slightly more useful than the second, in that it narrows down where the error is happening.

I can add around a dozen lines of code to make the debug message also include the file and line number of where `DB_query` was invoked. It makes it even easier for the developer, as he won't have to hunt around grepping for label strings, many of which are very much duplicated anyway.

>  ...it seems to me that we have a "monster pull request" 

That's the unfortunate consequence of a design decision which I respectfully don't agree with, namely putting queries everywhere. There's not much that can be done about it now.

> ...in order to slightly reduce the code size...

I think that reduction in translation strings might be the best outcome here, even more so than the reduction in lines of code

> ...with no previous complaints from any users that I recall

@pakricard did complain, that's why I started this work. See the comment he left on my previous commit, as linked above

gggeek:
> I can add around a dozen lines of code to make the debug message also include the file and line number of where DB_query was invoked

Done. Change pushed to this branch and is part of this PR

timschofield:
> > I can add around a dozen lines of code to make the debug message also include the file and line number of where DB_query was invoked
> 
> Done. Change pushed to this branch and is part of this PR

Thanks Gaetano I will check it and merge it later on.

timschofield:
> 
> > ...with no previous complaints from any users that I recall
> 
> @pakricard did complain, that's why I started this work. See the comment he left on my previous commit, as linked above

Ricard has been an active user and developer of webERP for 16 years, and yesterdays comment (that I had missed) is the only one I can find on this subject.

gggeek:
Thanks 4 merging

pakricard:
Hi:

As a rule of thumb, I think reducing lines of code makes sense. In this case ($DbgMsg), it makes all the sense as it is information only useful to the developer, and it is (was) inconsistent, as some DB_query had it, some not. Some $DbgMsg were defined and not used and some DB_query calls with a not defined $Dbg;sg, etc.

I prefer this debug info being coded just one, and complete, that scattered with (some) inconsistencies. I think it is a good step forward. ;-)

As I worked with it, I had to search for a string similar to the debug message and then try to locate the offending SQL. Now, should be easier as we will have the script name and line number. Direct to the culprit!

Attacking this has been on my to-do list for a long time, only to get dusty as urgent issues get the priority. I have some other "code simplification ideas", but we'll deal with them post v5 ;-)

Thanks to both of you 
R

PR #564, opened 2025-08-07
Explain purpose of dump_database.sh script

dalers:


timschofield:
I think that the new installer needs the database structure needs to be in separate sql files for each table

PR #567, opened 2025-08-08
fix usage of mysql extension, which is not in php8; remove one unused db dump

gggeek:


gggeek:
PS: I will update this PR to:
- not delete the mysql connector
- make sure the automatic conversion of the 'mysql' string in config.php to 'mysqli' only happens if the mysql extension is not available

gggeek:
Ready for review and merge.

PR #580, opened 2025-08-13
Clean up of $Debug instances, and repeated code about managing error codes from DB. 

pakricard:
Managing error codes from DB should be done in DB_query() only for consistency.

Some $Debug == 1 were development style messages to verify the SQL, etc. They should not be in a production code. Seem like leftovers from development times.

Some DB_query() used the 5th parameter to false, forcing DB_query fucntion to not execute the error trapping, and executing the error trapping in the script. It produced repeated an inconsistent code. Now DB_query() always does the error trapping. 

Some error trapping in the code contained DB_Txn_Rollback(), even if called with the 4th parameter as false. Simplified to call to DB_query() with 4th parameter as true and let DB_query() handle the rollback properly.

Always Set the DB_query to use error trapping, not using the 5th parameter (default = true). DB_query() should always do the error trapping. Why would we don't want to know an unexpected error occurred? Should we delete the 5th parameter from DB_query()?

In SuppPaymentRun.php, line 190 there is a DB_Txn_Commit() and just after that, we check if (DB_error_no() !=0). 
I am not sure if it is an overkill, or we should really improve function DB_Txn_Commit() and take care of the errors it produces, not only in one random file. Code in SuppPaymentRun.php still not removed. 

R

P.S.: My first PR. Not entirely sure if it's done correctly or not ;-)

### DIFF
--- a/AgedDebtors.php
+++ b/AgedDebtors.php
@@ -242,20 +242,8 @@ if(isset($_POST['PrintPDF']) or isset($_POST['View'])
 				holdreasons.reasondescription
 				HAVING ABS(SUM(debtortrans.balance)) >0.005";
 	}
-	$CustomerResult = DB_query($SQL,'','',False,False); /*dont trap errors handled below*/
-
-	if(DB_error_no() !=0) {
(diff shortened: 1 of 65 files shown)


pakricard:
I

PR #581, opened 2025-08-13
Add global declarations for when using variables defined in other scripts

gggeek:
To avoid a huge PR, this touches files named A to F.

Also:
- use more uniform spacing for the includes block at top of file 
- remove the name of the file from comment blocks at top of files
- fixes bugs due to use of undefined vars, such as $ErrMsg
- adds `@bug` comments where there is use of a missing var that I could not really grok


gggeek:
Closing, as no consensus was reached, and this branch has gone stale (too many conflicts now)

PR #585, opened 2025-08-15
Use custom translation method `__(` instead of `_(`

gggeek:
I know that this is big, but there is no way to make this change without touching all the files which have a translation call.

The benefits were discussed in some tickets (or discussion), namely: this will make translations work even when the user-preferred locale is not installed in the operating system


timschofield:
Hi Gaetano, I have been trying to merge this, but am getting a ton of conflicts. Can you redo it against the latest code?

Thanks
Tim

gggeek:
Closing in favour of same changes re-done in PR #587

PR #586, opened 2025-08-15
Check for and remove UTF BOM character sequence if present.

paulcsf:
In Z_ImportGLTransactions.php check for and remove UTF BOM character sequence if present.


paulcsf:
There is a (more?) complete implementation in the old PHPExcel files, function _skipBOM().

Issue #588, opened 2025-08-17
Deprecated error after PR #587


pakricard:
Hi @gggeek :

After PR https://github.com/timschofield/webERP/pull/587 I get this error:

Deprecated: Using ${var} in strings is deprecated, use {$var} instead in /weberp/vendor/gggeek/polyfill-gettext/src/T.php on line 552
same on lines 553, 555, 560,561,563

R

dalers:
Hi @pakricard why did you close this issue? What was the resolution?

Fwiw, my apache log also has errors related to polyfill-gettext and I'm wondering if there is a general issue related to our (@gggeek)   polyfill-gettext.

On my server I get these errors also:

`[Fri Aug 15 19:59:13.289952 2025] [proxy_fcgi:error] [pid 65256:tid 28331151990272] [client 174.0.43.39:53186] AH01071: Got error 'PHP message: PHP Warning:  Function T::setlocale called with LC_ALL category, but in emulated mode only messages will be translated. Use LC_MESSAGES instead in /usr/local/www/weberp/vendor/gggeek/polyfill-gettext/src/T.php on line 458'
[Fri Aug 15 20:01:05.529313 2025] [proxy_fcgi:error] [pid 65256:tid 28331151990272] [client 174.0.43.39:53204] AH01071: Got error 'PHP message: PHP Warning:  Function T::setlocale called with LC_ALL category, but in emulated mode only messages will be translated. Use LC_MESSAGES instead in /usr/local/www/weberp/vendor/gggeek/polyfill-gettext/src/T.php on line 458', referer: https://weberp.dalescott.net/install/index.php?Page=5`

gggeek:
About the 1st error reported: weird that I did not get it on my pc, nor got it triggered via the unit tests. If it still does happen, the correct place to report a bug, a PR and a fix is the upstream polyfill-gettext repo.

About the 2nd issue: I will have to check, as that warning is triggered depending on how the polyfill api is called. And I did not think that I had made weberp call it in a wrong way. Can you tell me on which pages you get it? All of them?

dalers:
Hi @gggeek 

> ...Can you tell me on which pages you get it? All of them?

I can't say what triggers the warnings but will watch the log more carefully in the future and update when I know more.

timschofield:
I see this, but it only appears the first time that page is called after a restart

[...]

PR #589, opened 2025-08-18
bump polyfill-gettext; make sure we alway load session.php before…

gggeek:
translating text; formatting improvements for include calls


gggeek:
Fixes #588

PR #602, opened 2025-08-22
Move from manual inclusion of 2 files to using static class methods

gggeek:
The benefits of this approach are:
- less usage of `include` calls in code - relying on class autoloading is better, avoids double inclusions
- less usage of global variables - helps some dumb IDEs 😉 

The LanguageSetup.php file is unchanged - I just removed commented-out code.

The original LanguagesArray.php and CurrenciesArray.php are still left in, and will behave as before, but trigger a warning if used. I think this is the best solution for keeping BC and nudging users to modify their own code if they did any customisations...


timschofield:
Using include files was a conscious decision as we considered it makes the code more readable. I'm not sure I see any advantages to changing this.

gggeek:
> Using include files was a conscious decision as we considered it makes the code more readable. I'm not sure I see any advantages to changing this.

I knew this could be controversial...

The thing, as far as I see it, is that the current way of using includes, and global variables, is quite messy.

Not only it breaks parsing for some IDEs, but also:
- it leads to more code lines. Every include/require call transformed into "autoload magic" is one less line of code. Esp. important when refactoring to introduce eg. usage of `$RootPath` or `__DIR__` everywhere
- it leads to code inconsistency: there is an abundance of require/include/include_once/require_once in the current codebase, not a single standard in use. Ditto for using a prefix to the path of the included file
- it can lead to involuntary double inclusion of the same file (seen many times, a few corrected in this PR)
- included files behave differently when they are included from the global scope and when from within a function. This is the reason why there were already `global $var` declarations here and there before I tried to introduce more
- included files might have side effects or not. When a file with a class definition is included, it only has the class definition in it, and no side effects
- it makes it easier to introduce subtle bugs by having name clashes in global variables
- it makes code less easy to test
- it makes code _less easy_ to read, at least to me. When using an IDE, I can ctrl+click on a method call, and be taken to the declaration. With global vars, it is generally up to me to figure out where those are defined (I think we already settled the discussion about using an IDE, whichever one, being a fundamental prerequisite)
- is generally considered a bad practice and might deter younger coders from joining development

dalers:
> I'm not sure I see any advantages to changing this.

I also don't see the advantages but willing to be educated.  Are you willing to climb the hill and still die at the top? I discounted arguments with "leads to", "can lead to", "might have", "at least to me", "considered bad practice" and focussed on "easier" aka more efficient feature development and debugging, likely using an IDE and likely AI guided. If that can be shown, I would be a believer. However, because there is no pain felt now, it's hard to drum up much enthusisiam for the work.

Is there a prototype aka MVP aka proof-of-concept POC that would demonstrate the difference? How do you feel about creating a side-by-side video/youtube demo with your favorite IDE and current code on the left and POC for proposed changes on the right?

timschofield:
Hi @gggeek and @dalers 

The fact that we don't follow latest PHP fashions in our coding is not news, it's been pointed out many times over the last 20+ years :)

The aim has always been to create code that fellow accountants can pick up easily and change. I think the history of the project shows that we have been largely successful in this aim. I also feel that sticking to our methods, rather than changing with the latest fashion is one of the reasons for our longevity. I have been involved in other projects over the years that have died after constant rewrites to keep up with current "best practices" and this has led to developer fatigue. Personally I would rather add new functionality to webERP than rewrite the code to meet the latest fashion.

As for changing the code to suit an IDE that seems like the tail wagging the dog to me. I use various command line tools to do my development as I have always done so get no benefit from this. I really appreciate the efforts @gggeek is putting in to the project, and I have merged several pull requests today already. I just don't see any advantage to us with this one.

Thanks
Tim

gggeek:
> Is there a prototype aka MVP aka proof-of-concept POC that would demonstrate the difference? How do you feel about creating a side-by-side video/youtube demo with your favorite IDE and current code on the left and POC for proposed changes on the right?

This is the 2nd PR which I propose on this "topic", broadly defined as "getting rid of include calls and global variables". I have exposed the expected benefits both in the comment above and in lengthy discussions in the "big refactoring" GH discussion (#453).

I am not sure that recording a video would bring much more to the table...

As we already discussed, I am the only one of the currently active developers using PHPStorm, so showing how much my IDE gets smarter when cross-file global variables are eliminated and class methods are used instead does not seem to be a point that would convince anyone.

The ease of navigating code seems to be pretty much a matter of personal taste - with Tim strongly in the camp of "following includes is easier". I suspect he also dislikes classes and favours functions (but don't quote me on that).

The ease of refactoring due to less lines of code, the reduced amount of bugs, etc.. are also hard to demonstrate - either we are talking about small/cosmetic fixes, or about introducing features that would require much more development than a PoC but a proper hardcore refactoring, and, most likely, would not be rated highly in any case, due to the perceived increase in "complexity" being greater than the added value.

I could f.e. refactor the current DB connector classes in a way that splits the definition of the functionality from the act of creating the database connection, thus decoupling the DB from the session, and making it easy to implement a non-session-based API for weberp. But in past discussions, there was little to no appetite for that...

Let me think about this and see if there's anything which would be a clear win...

timschofield:
> I suspect he also dislikes classes and favours functions (but don't quote me on that).
> 

You are getting to know me :)

dalers:
> I also feel that sticking to our methods, rather than changing with the latest fashion is one of the reasons for our longevity.

While I would agree webERP's goal is to target SME's with a champion who is both business and techical (or vice versa) (and possibly or likely an accountant), adopting general developments in how software is developed would benefit the project _**so long as they do not compromise the goal**_. Taking @gggeek at his word, new community members will find the code less confusing and arbitrary and more compatible with tools currently commonly used in industry.

> recording a video

a picture is worth a thousand words, multipled by frames per second....

> showing how much my IDE gets smarter when cross-file global variables are eliminated and class methods are used instead does not seem to be a point that would convince anyone.

@timschofield may never be convinced. @pakricard is probably already a believer. I'm a sceptic because "it's" working for the project now and I don't have the personal experience to know otherwise. If you can show me (instead of telling me), I'm much more likely to become a believer.

> are also hard to demonstrate

Sounds like the cost of quality...  you're either a believer or not (or is that religion ;-)).

> in the camp of "following includes is easier".

You say there is an actual camp of these people? This would add weight to the "it works now, don't break it" option as "following includes" isn't total heresy.

timschofield:
> You say there is an actual camp of these people? This would add weight to the "it works now, don't break it" option as "following includes" isn't total heresy.

It is still common in older projects to include files. I refer again to Wordpress that was started at more or less the same time as webERP. They use include/require throughout their codebase.

Tim

gggeek:
> I also feel that sticking to our methods, rather than changing with the latest fashion is one of the reasons for our longevity. I have been involved in other projects over the years that have died after constant rewrites to keep up with current "best practices" and this has led to developer fatigue. Personally I would rather add new functionality to webERP than rewrite the code to meet the latest fashion.

TBH, I do not think that my proposals could be described as trying to "meet the latest fashion".
All the PRs sent are based on well-know, long established practices, which are not likely to go away any time soon.

Extensive usage of Composer to manage dependencies: Composer was introduced around 2013.

OOP programming: the most widely used paradigm for coding in PHP, since version 5.

Support for autoloading classes: in PHP since version 5. Got improved over time.

I could in fact be advocating for far more radical changes, such as f.e. introduction of a framework such as Symfony or Laravel, introduction of an ORM, usage of the MVC pattern and a templating language plus Dependency Injection Controller, but I personally consider that those are indeed subject to the wear-and-tear of the latest-fashion and would bring to the table enough complexity to make the project non-maintainable by non-professional developers.

> Personally I would rather add new functionality to webERP than rewrite the code to meet the latest fashion.

Sorry for repeating myself, but my pow on this is that the current codebase actually makes it hard to introduce new functionality, _precisely_ for the way it is architected, not the opposite. The only way to be effective at that is to be already knowledgeable with existing code.
Ex:
* using 'include' statements does not show to the developer what the contract is between including and included files, ie. which variables are supposed to be passed along, let alone their type
* in the same way, the code is not very discoverable, which means lots of "simple" functionality is implemented in the many controllers via copy-paste, and that ends up with myriad variations, many of which are slightly wrong. See f.e. the mess which was the usage of variables $RootPath and co. to display links to other pages/images/css/js files. If there was a function 'buildUrl()' available to encapsulate the prefixing of $RootPath, that would not have happened
* the fact that the sql statements are not grouped in a data-accessing-layer but hardcoded in every controller makes it exceedingly hard to refactor the db schema
* the fact that the SESSION is used everywhere makes it hard to develop cli scripts. Those are necessary for data import/export, to avoid issues with timeouts and OOM

> It is still common in older projects to include files. I refer again to Wordpress that was started at more or less the same time as webERP. They use include/require throughout their codebase.

WP is _universally_ considered a horror show of a codebase, which is hampered in its evolution precisely because it was badly designed in the 1st place. It keeps chugging along just because it is extremely well tested to prevent breakages and has a huge base of developers. And even then, breakages of plugins are a daily occurrence.

gggeek:
@dalers not a video, sorry, but a practical example - lots of words:

see the change at https://github.com/timschofield/webERP/pull/602/files#diff-68e9ac14b40e479ca26784a70d1e2dddded6eba1c810d20b3f954d5b9f54595eR4

In this case, the code is using `include('includes/header.php');` as a "way" to display the html header.

The contract for doing that is not defined anywhere - one has to open 'header.php' and read it in its entirety to figure out which variables it expects to be defined.

Once the developer writing the 'SuppLoginSetup.php' controller had figured out that there are 3 variables commonly used to drive the display of the header, namely `$Title`, `$ViewTopic` and `$BookMark`, he still made the mistake of defining 2 of the 3 variables _after_ the inclusion of 'header.php'
I am 99% sure that this is an error, and have corrected it in the PR.

_BUT_ there is no easy way for me to be 100% sure this was not intentional. Maybe the original dev actually wanted to reset the value of those vars after the display of the header, because they are reused further down the page?
The only way to be sure is to read the rest of the code of the webpage, and all the other php files it includes, one by one. And there is no way that an IDE would ever be able to flag this as an error or even a code smell

One could of course develop a small bash/php script that checks every php file for that specific type of error by using 'strpos', but it's a lot of work for a one-reuse script... it could maybe be implemented as a rule for phpstan - but who has time to introduce phpstan if we're not ever working on code quality? And how many custom phpstan rules should we write, to catch all possible mistakes in using 'include' calls?

What if we replaced instead the 
```include('includes/header.php');```
line with
```webERPPage::displayHeader($Title, $ViewTopic, $BookMark)```
?

It would be immediately evident when someone is not passing in the correct arguments.

And the IDEs would be able to give tooltips for the 3 arguments, with their type and a short description, at the time that that line of code gets written...

gggeek:
>> I suspect he also dislikes classes and favours functions (but don't quote me on that).
>
> You are getting to know me :)

Just for the record: I am not advocating for the usage of "fancy" OOP concepts, such as subclassing and polymorphism. Mostly, I'd like to see classes introduced as a way to organise code, so that it's not a thousand functions spread out (defined) everywhere, but grouped in a more manageable hundred classes with a hundred methods each.

As a side note, there are already quite a few classes in webERP - I'm not advocating for introduction of something entirely new.

pakricard:
As a general goal. I think we should go towards a reduction of numbers of lines in webERP.

webERP might seem "easy to read" to a novice, but the fact that many bits are repeated, with slight variations (mainly due to the coder, not to logical differences), makes it difficult to understand as a whole. I systematically failed to train a PHP developer in my team.

webERP has been around for 20 years, but besides some facelift, I think it is time for a more profound code quality improvement.

I do see advantages in removing as much manual added includes as possible.  I always need to open them, check and remind the variable names, then use them wisely (a.k.a. don't touch much) 

Autoload is smarter than us ;-) and will load whatever is needed. Fewer lines of code to maintain, more efficient code.

I do see many advantages in encapsulating repeated code here and there in functions. as example, I have some UI functions like FieldToSelectMultipleLocations(), FieldToSelectMultiplePeriods(), FieldToSelectOneLocation(), etc. to lower drastically the number of lines of code needed when building field sets and increasing the consistency. It's not the time to discuss this now, let's wait until post v5)

I see certain advantages in the use of OOP. We have some classes here and there, but not extensively used. It forces the rewriting of a lot of code, and it's not the moment now to discuss this.

> Once the developer writing the 'SuppLoginSetup.php' controller had figured out that there are 3 variables commonly used to drive the display of the header, namely `$Title`, `$ViewTopic` and `$BookMark`, he still made the mistake of defining 2 of the 3 variables _after_ the inclusion of 'header.php' I am 99% sure that this is an error, and have corrected it in the PR.
> 
> _BUT_ there is no easy way for me to be 100% sure this was not intentional. Maybe the original dev actually wanted to reset the value of those vars after the display of the header, because they are reused further down the page? The only way to be sure is to read the rest of the code of the webpage, and all the other php files it includes, one by one. And there is no way that an IDE would ever be able to flag this as an error or even a code smell

I would bet those are inaccuracies that have been in the code for years, as some developer inherited and copied bits and bobs from other scripts (I was one of those. I try to improve now ;-))

R

timschofield:
> Sorry for repeating myself, but my pow on this is that the current codebase actually makes it hard to introduce new functionality, _precisely_ for the way it is architected, not the opposite. The only way to be effective at that is to be already knowledgeable with existing code. Ex:

Well I would say that history is on my side on this one. For instance when Mark Yeagers' company needed an MRP he wrote and submitted it as his first contribution without even asking for advice, it came fully working out of the blue - At the time we were the only open source ERP to have a working MRP as it's very complex.

Likewise when Andrew Galuski's company needed QA functionality he wrote it and submitted it without needing advice.

In fact the history of webERP is littered with contributions such as this where an organisation needed some functionality so they just wrote it, because webERP makes it easy to do.

>     * the fact that the sql statements are not grouped in a data-accessing-layer but hardcoded in every controller makes it exceedingly hard to refactor the db schema

My experience is that having the SQL where it is actually used in the code makes it more readable. It certainly makes it easier to figure out any problems.

> 
>     * the fact that the SESSION is used everywhere makes it hard to develop cli scripts. Those are necessary for data import/export, to avoid issues with timeouts and OOM
> 

As an accountant I recoil in horror at the idea of webERP being updated by command line scripts rather than using the interface. Accounting systems depend on an audit trail being maintained. It is this that makes the Utilities scripts so dangerous as I have been describing in another thread.


> WP is _universally_ considered a horror show of a codebase, which is hampered in its evolution precisely because it was badly designed in the 1st place. It keeps chugging along just because it is extremely well tested to prevent breakages and has a huge base of developers. And even then, breakages of plugins are a daily occurrence.

Wordpress is very widely used so whatever they are doing seems to work pretty well :)
I only invoke Wordpress because I think Dale was under the impression that only webERP was stupid enopugh to develop in this fashion, whereas it is actually still common.

Tim

timschofield:
> Once the developer writing the 'SuppLoginSetup.php' controller had figured out that there are 3 variables commonly used to drive the display of the header, namely `$Title`, `$ViewTopic` and `$BookMark`, he still made the mistake of defining 2 of the 3 variables _after_ the inclusion of 'header.php' I am 99% sure that this is an error, and have corrected it in the PR.

This was me I think. $ViewTopic and $BookMark are there to guide the help system to the correct place in the manual if the user clicks on the manual icon in the top right. In this case the manual page isn't written so the user would just get a default page, as they would if the variables aren't declared. I seem to have done it right in the other 450+ scripts, but you are correct this needs correcting - but not as part of an unrelated PR :)

Thanks
Tim

gggeek:
> Well I would say that history is on my side on this one. [...snip]

I suppose you are aware of the anecdote of the call-center which though that all systems were go because they received no support calls for one whole day, when in fact it was their phone lines which were down?
Which is to say that it is generally hard to assess how many opportunities you are missing...

> My experience is that having the SQL where it is actually used in the code makes it more readable

It makes maintenance and evolution harder. It allows for many small variations to creep in. Those 2 factors outweigh the readability imho.

> Accounting systems depend on an audit trail being maintained. It is this that makes the Utilities scripts so dangerous as I have been describing in another thread.

Using command line scripts instead of web pages has nothing to do with accountability.

In fact, the only way to make sure that an audit trail is maintained is to force every access/change to data to go through a data-access-layer which takes care of the auditing, and forbid direct execution of sql from anywhere else (which is a valid rebuttal to sql-in-controllers, btw...)

Also, making dangerous and long-running changes only available for execution via the cli goes hand in hand with increased system security, as generally such access is restricted to admins. And it is still possible to have a username/password be required to run the scripts, so that the accountability trail and auth checking are preserved.

Having said that, I think that we're again going a bit OT for the discussion on this PR. I will restrain myself from further comments.

dalers:
> Accounting systems depend on an audit trail being maintained.

I would say _**any**_ action causing a write to the database  _**must**_ be recorded in the audit log, regardless of whether the result of a logged in user in the GUI or a CLI script included with webERP. It may also be desirable for running a report to be logged in the audit log for forensic purposes, e.g. so that there is a record of a sales person running an all-customers report the day they resigned.

dalers:
> I seem to have done it right in the other 450+ scripts, but you are correct this needs correcting

That's a LOT of work! (I might even speculate a challenge to complete before more urgent issues arise as they typically will).

This is one reason coding practices have evolved - to make this sort of simple mistake unlikely or ideally impossible to occur.

timschofield:
> That's a LOT of work! (I might even speculate a challenge to complete before more urgent issues arise as they typically will).
> 

It was done many years ago when I was a much younger man ;)

I have fixed the incorrect one!

gggeek:
> I would say any action causing a write to the database must be recorded in the audit log, regardless of whether the result of a logged in user in the GUI or a CLI script included with webERP. It may also be desirable for running a report to be logged in the audit log for forensic purposes, e.g. so that there is a record of a sales person running an all-customers report the day they resigned.

In my experience, the best format for the audit log is similar to: timestamp, user id, action, affected object id. Extra details are only useful in specific cases.

The "action" might something like 'item/delete', 'user/login', etc... As such, it does not lend itself well to being inferred from the log of db queries, because a) some actions might not result in any query at all, b) some actions might require more than one query, and c) there might even be an n-to-m mapping between the two.

The webserver access log is not a good candidate for this, unless it is augmented with the current-user, action and target-object-id, which is unwieldy to do (also: what about many actions happening in 1 webpage? and actions executed from the cli?).

The full log of db queries is in general both too verbose, and, as stated above, not necessarily easy to match to 'actions' taken by users. It can also be generally activated on the side of the db-server, which has the benefits of 1) not having to be coded in the weberp (fewer lines of code!), and 2) being useable as a complementary tool, in case someone sneakily modifies data outside of weberp by direct db access.

gggeek:
> I have fixed the incorrect one!

Thanks.

While at it, it might be worth removing the useless (duplicate) inclusion of LanguagesArray.php from CustLoginSetup.php line 9, SuppLoginSetup.php line 9, includes/Login.php line 5, install/index.php line 50

timschofield:
The audit trail in webERP is primarily there for the accountants to see what changes were made to each table, when, and by whom. Having a clear trail of this is very important. I have had the audit trail looked at by the auditors of many organisations around the world, and it has met with their approval. If someone has a better way of producing the same information in the same format, then great, otherwise it's best left as is.

Tim

gggeek:
PS: slightly related another minor issue with bad/useless usage of 'include' calls fixed in PR #610

timschofield:
If I was a novice user who is reading a script such as AnalysisHorizontalPosition.php and I came across the line

include_once('includes/CurrenciesArray.php');

then it is obviously including the file includes/CurrenciesArray.php at this point, and I can quickly open that file in my editor of choice and see what it is doing.

However if I came across the line

$CurrencyName = \webERP\CurrencyManager::getCurrencyNames();

then it isn't obvious at all how I could view that in my editor. Yes I could probably work it out, but what is the point of changing something that is clear and obvious to something that appears to be deliberately obfuscated?

This PR makes the code less readable while adding no functionality to webERP and I cannot see how it could be changed to improve that. Unless someone can find a way of doing that I am minded to close it.

Tim

gggeek:
@timschofield feel free to close this PR - it seems we're never going to agree here, and you are ultimately the tech lead of the project.

You have failed to convince me though, even with your latest example.

To me, anyone capable enough to contribute to webERP or even just tweak it to his/her own needs, has to have a basic understanding of the PHP programming language.
And classes and class autoloading are basic features of PHP, having been there for a long time.

The usage of a static class method to return an array of data does not seem to require deep understanding of OOP intricacies - the main difference with the way this is currently done is figuring out where the 2nd source code file is.

I do not believe that any novices will be reading and editing php code with a plain text editor - in fact they are more likely to be using a (free and easy-to-set-up) IDE, as that makes it much easier to navigate existing code and write new code (syntax highlighting and all).

And, every IDE which supports PHP also has integration with the Composer autoload mechanism, meaning that the simplest way to find out where the code for `CurrencyManager::getCurrencyNames` is is to ctrl+click on it.

Or just ask Copilot...

For the more technically inclined, it's not very hard to understand that, given the fact that we are using the PSR-4 class naming standard, and that the `/src` folder is mapped to the root of the `\webERP\` namespace (this is all defined in composer.json), class ` \webERP\CurrencyManager` must be defined in file `/src/CurrencyManager.php`.

Or they could just grep for `class CurrencyManager` and find the matching file.

This might not be second nature atm to all the developers involved, but it would become quite evident if we adopt this principle and start moving all existing classes to using PSR-4...

gggeek:
> what is the point of changing something that is clear and obvious to something that appears to be deliberately obfuscated

I have gone to great lengths to explain in detail the expected benefits, and the issues inherent to the the current scheme, in the previous comments in this PR. I do not think I can do much more to try to convince you :-(

timschofield:
> You have failed to convince me though, even with your latest example.

I'm sorry to hear that.

> 
> To me, anyone capable enough to contribute to webERP or even just tweak it to his/her own needs, has to have a basic understanding of the PHP programming language. And classes and class autoloading are basic features of PHP, having been there for a long time.

A basic understanding of the PHP language - yes
We have always used classes where we have deemed they are appropriate, but not just for the sake of it.

> 
> The usage of a static class method to return an array of data does not seem to require deep understanding of OOP intricacies - the main difference with the way this is currently done is figuring out where the 2nd source code file is.

Exactly.

> 
> I do not believe that any novices will be reading and editing php code with a plain text editor - in fact they are more likely to be using a (free and easy-to-set-up) IDE, as that makes it much easier to navigate existing code and write new code (syntax highlighting and all).

Not my experience at all. Most people who want to start "tweaking" the code do not start with downloading an IDE, and then devoting time to learning how to use it. Most see a text file, and so load up their text editor and start changing it.

> 
> And, every IDE which supports PHP also has integration with the Composer autoload mechanism, meaning that the simplest way to find out where the code for `CurrencyManager::getCurrencyNames` is is to ctrl+click on it.
> 
> Or just ask Copilot...
> 
> For the more technically inclined, it's not very hard to understand that, given the fact that we are using the PSR-4 class naming standard, and that the `/src` folder is mapped to the root of the `\webERP\` namespace (this is all defined in composer.json), class ` \webERP\CurrencyManager` must be defined in file `/src/CurrencyManager.php`.
> 
> Or they could just grep for `class CurrencyManager` and find the matching file.
> 

Pretty sure I said earlier today that of course it can be worked out. My point is that currently the code is obvious. Having to start using Copilot, or grep makes something that is currently easy, much harder for no benefit to the end user.

PR #607, opened 2025-08-23
use a prerendered image with transparent corners for company logo

gggeek:
while at it, rename the fallback image used as company logo when the generation fails


dalers:
I didn't actually test the code but viewed the image and agree with merge.

PR #617, opened 2025-08-26
make start of web pages code more uninform, p

gggeek:


timschofield:
PDFAck.php breaks with these changes as it requires DomPDF.

Maybe others, PDFAck.php was first on the list

Tim

gggeek:
You are right.
Something weird has gone on with PDFAck.php - in all files changed I just shuffled code around, removed comments, changed whitespace and removed duplicate statements, but here it seems the start of file was changed in a different way... bad copypasta or git confusion?
I will fix.

gggeek:
ps: this might have happened because you changed that file too, after I created the branch. I will try to merge master into this branch and see if I get any conflicts

gggeek:
Indeed that was the cause. It found that and another file to fix.

gggeek:
Fixed.

Btw - if you include `session.php`, you do not need to require manually also `vendor/autoload.php`, as session.php takes care of that

timschofield:
> Indeed that was the cause. It found that and another file to fix.

Ah, that was changed by github trying to resolve the conflicts that was there in your original PR

timschofield:
PO_Header seems to have lost these declarations as well.

$ViewTopic = 'PurchaseOrdering';
$BookMark = 'PurchaseOrdering';

gggeek:
> PO_Header seems to have lost these declarations as well.

Correct. Sorry about that - this time developer distraction. Fixed now

PR #618, opened 2025-08-26
make start of web pages code more uninform, q-r

gggeek:


timschofield:
I think that

$ViewTopic = 'SalesOrders';
$BookMark = 'RecurringSalesOrders';

have also been lost from RecurringSalesOrders.php

Tim

PR #627, opened 2025-08-26
lowercase Class and Var

gggeek:


timschofield:
This has some conflicts, can you check and redo?

tim

gggeek:
Fixed

Issue #487, opened 2025-06-26
Clean up usage of `RootPath` and `PathPrefix` global variables


gggeek:
It seems that the current code is slightly confused as to what those 2 variables mean, and how they should be used.

### RootPath

denotes the "root URL" of the website, as seen by the end user's browser

* it is used as part of URL building
* it does not contain hostname or port
* current issues:
    * it is either determined dynamically, based on `PHP_SELF`, or set to a fixed value in `config.php`. But the value of `PHP_SELF` varies, depending on where the starting php file is located, ie. it will be one-directory deeper if it is within `/intall` or `/reportwriter`. There is an attempt in the code to work around that using global variable `$DirectoryLevelsDeep`, but that won't work when $RootPath is set to a fixed value

Can we improve this, making the code f.e. automatically adjust the `$RootPath` based on alaysis of `__DIR__`?
We could decide that a default value of `auto` means that it is to be auto-calculated (I can send a PR for that)

### PathPrefix

denotes the "root dir" of webERP on disk

* it should not be used to build links
* it is easy to determine, in `session.php` or one of the files which are included everywhere
* it should not be set/changed by any other files
* it should always be used for any `include` \ `require` call in our code

If in agreement, I can work on this

timschofield:
Agreed, this has got very confused over the years.

gggeek:
> Agreed, this has got very confused over the years.

Any strong opinion on using absolute paths instead or relative ones?

gggeek:
First cleanup pass done, in PR #488 

Left:
- make sure all include calls use PathPrefix
- optionally, rename RootPath to UrlPrefix

timschofield:
> > Agreed, this has got very confused over the years.
> 
> Any strong opinion on using absolute paths instead or relative ones?

Sorry, I have been out and only just seen this. On balance I would say relative paths, but that's not a strong opinion

gggeek:
> On balance I would say relative paths, but that's not a strong opinion

Too late! 🤣 

I went with absolute paths because atm it's a bit of a mess to figure out the directory of the currently executing script. Some scripts (dashboard applets) even have provisions to be run either via an `include` call or executed standalone as top-level webpage! 😰 .

Using __DIR__ on the other hand is non-controversial, as it always refers to the file it is written in, not the original includer.

pakricard:
> Any strong opinion on using absolute paths instead or relative ones?

I would say absolute paths.

[...]

PR #637, opened 2025-08-27
always require session.php instead of including it…

gggeek:
...and use absolute of relative path for inclusion

also:

fixes to edi_config.php
remove one redundant 'global' declaration
fixes to glprofit_loss.php for when included from another script
a couple leftovers from the cleanup of start-of-page includes

### DIFF
--- a/AccountGroups.php
+++ b/AccountGroups.php
@@ -2,7 +2,7 @@
 
 /* Defines the groupings of general ledger accounts */
 
-include('includes/session.php');
+require(__DIR__ . '/includes/session.php');
 
 $Title = __('Account Groups');
 $ViewTopic = 'GeneralLedger';
--- a/AccountSections.php
+++ b/AccountSections.php
@@ -2,7 +2,7 @@
 
 /* Defines the sections in the general ledger reports */
 
-include('includes/session.php');
+require(__DIR__ . '/includes/session.php');
 
 $Title = __('Account Sections');
 $ViewTopic = 'GeneralLedger';
--- a/AddCustomerContacts.php
+++ b/AddCustomerContacts.php
@@ -2,7 +2,7 @@
 
 /* Adds customer contacts */
 
-include('includes/session.php');
+require(__DIR__ . '/includes/session.php');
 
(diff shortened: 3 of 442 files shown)


timschofield:
Can you explain what problem is being solved by this, and why that problem hasn't arisen in the past 22 years?

Tim

gggeek:
1. using `require` instead of `include` is a safety measure - if, by php misconfiguration or a temporary editing mistake the 'session.php' file is not available for inclusion, halting all processing is the best choice, as it prevents the code from doing unintentional damage, and from showing to the end users error messages which might contain information potentially useful to an attacker
2. using the full path to the included file (via `__DIR__`) makes the code slightly more robust, as it does not depend any more on the php.ini setting 'include_path' or any stray call to 'set_include_path'. I think it might also be marginally faster, and help with php opcode cache and timestamp validation or realpath cache (I forgot the details)

I though that moving all includes to using absolute paths was agreed upon - it was discussed in some other ticket.

And, in my mind at least, this work is part of / related to ticket #487

I appreciate it is a big change.

gggeek:
ps: this is a good writeup about the small but interesting difference between using absolute or relative paths for include/require calls: https://cjhaas.com/2019/05/21/php-include-path-surprises/ - worth reading all the way to the bottom. 

TLDR:
* including a file using the syntax 'includes/something.php' makes php go through the include_path. More IO, hogs resources
* including a file using the syntax '../includes/something.php' has different behaviour when a script is run from the command-line
* the best solution is to always go for absolute paths

timschofield:
If somebody has removed your session.php file then the error messages being sent are the least of your problems :)

It seems we are changing 442 files for no real benefit when that line of code has worked fine for 22 years. Hopwever I will merge this PR as it also does no harm either.

Tim

gggeek:
> If somebody has removed your session.php file then the error messages being sent are the least of your problems :)

Lol true.

I was thinking more about some unintentional misconfig from the resident sysadmin/lead-developer rather than an attacker having already gained write access to the app's filesystem.

Thanks for merging!

PR #643, opened 2025-08-27
use $RootPath for all links, part 1

gggeek:


gggeek:
Part of the fixes for issue #487

PR #666, opened 2025-08-29
Rename GoogleTranslator.php to TranslationFunctions.php

gggeek:
benefit: move towards a more uniform naming standard for includes, and possibly eventually a better organisation of the defined/available functions


timschofield:
Like several scripts in the Utilities section this script is totally broken. It could be made to work again but requires TLC, so if you have time ..... :)

gggeek:
Well, renaming the file is is the 1st step in the long road to fixing it.

I am aware that many utilities need maintenance... eg. the scripts for dealing with translations seem to be quite broken.

pakricard:
I coded this 11 years ago, used for a while and stop using it because the translations (at that time) were terrible and all of them needed human correction. 

In these 11 years Google API might have changed 11 times ;-) and code was never updated, and no one complained, so I guess, absolutely no one is using it. 

Then, we should decide is it is really a useful feature we would like to keep, or we can toss and forget as obsolete.

My vote is to keep hidden for now and fix post v5 as it is obviously a non-essential feture.

timschofield:
I think that unless someone is prepared to take on the job of maintaining this on an going basis so that they monitor the Google API and whenever it changes update the script, then I would leave it hidden as Ricard suggests.

dalers:
> I would leave it hidden as Ricard suggests.

So we are all clear, what do you mean by "hidden"? Imho we need a way to leave files that are known to not work in the codebase without leaving the impression they do work and are just not accessible via the GUI.

How about moving scripts that are known to not work to a new "obsolete/" directory (or please suggest an alternate name if you prefer).

pakricard:
What I currently do is create a Security Token 9999 --> "webERP - Script not used" and assign all the scripts I don't want anyone to use to this Security Token.

Then, ensure that no “Access Permissions” aka user Role has access to this 9999 Security Token. So, the script can't be run by any user, but it is still there for reference. 

Moving it to an /obsolete folder does make sense too, like a Recycle Bin.

Any approach will suit me.

gggeek:
@dalers +1

I'd go for "unmaintained" rather than "obsolete" - it conveys more the idea that if someone wants to pick up maintenance, those could be brought back to common use

gggeek:
ps: it just occurred to me that renaming any php script which is accessed from end users might "break" the security policies which were applied to it.

Can anyone confirm that is true?

If it is,
- I will refrain from sending PRs which rename any files, except the "include" ones
- I will update the dev cookbook with a warning about this
- we will have to mention any moving/renaming in the "upgrade to v5" guide

timschofield:
Yes, if you rename a file that the user can access directly (not an included one) then the page security token assigned to it doesn't travel with it. In my experience the average ERP user pays attention to the menu description they are clicking on, and then what the screen says to them after they have clicked on it. They will rarely even notice the file name. Only rename stuff in extreme cicumstances.

Tim

dalers:
@timschofield 

> then I would leave it hidden as Ricard suggests.

Do you "hide" scripts the same way as @pakricard ?

> What I currently do is create a Security Token 9999 --> "webERP - Script not used" and assign all the scripts I don't want anyone to use to this Security Token.
> 
> Then, ensure that no “Access Permissions” aka user Role has access to this 9999 Security Token. So, the script can't be run by any user, but it is still there for reference.

timschofield:
Hi @dalers sorry for the late reply, I am on holiday this week so not supposed to be doing webERP stuff ;)

If I have complete control I just remove the scripts I consider they shouldn't be using entirely. Also as I have said before I remove the utilities modulecompletely. In the unlikely scenario I would need to run one of the scripts I would just upload it myself and run it.

Ricard is in a slightly different position to me in that it is his company. My sites belong to the organisations themselves so I cannot stop them from loading up any scripts they want and running them - though it may cost them if they want me to put right the damage!

Tim

PR #673, opened 2025-08-30
fix usage of `use` statement for barcodepack; add one comment

gggeek:
Notes:

1. "use" statements are not used to import composer packages, but to import into the current namespace classes which reside in another namespace. 

As such, `use gggeek\barcodepack` is wrong. It might not be generating warnings, but it also does nothing.

Instead, `use BarcodePack\qrCode` means that, anywhere in the php file, one will be able to use `new qrCode()` where it would have been necessary to use `new \BarcodePack\qrCode`.
In short, it avoids having to write tedious, unreadable code.

2. I did not test this page, but it seems that QR codes are generated twice each, without a clear reason?


timschofield:
The code I committed works, and generates the correct QR code. I haven't tested your changes, have you ensured that it still produce the same output as mine?

Tim

gggeek:
> I haven't tested your changes, have you ensured that it still produce the same output as mine?

No 😓 

I will try to produce a work order out of the demo db. I don't know if that will be an easy process or if there is a lot of other data to be created first...

timschofield:
> No 😓
> 
> I will try to produce a work order out of the demo db. I don't know if that will be an easy process or if there is a lot of other data to be created first...

Have you tried running this script yet? I am conscious that my code worked, and have a note to go back and improve it but would rather put your changes in first or it will get messy

Tim

gggeek:
Sorry, I have been busy working on improving the installer (a prerequisite for me to be able to carry out any further testing of my changes, as I atm have no set up of webERP with a database filled with sufficient data to do anything meaningful), and that spiralled into creating the test suite to make sure that my changes did not break the installer.

Turtles all the way! 🤣

I will prioritise testing this change.

timschofield:
It would help me greatly if you held back your commits for a script until you have had the opportunity to run that script.

timschofield:
This change is still waiting for testing after 3 weeks?

gggeek:
> This change is still waiting for testing after 3 weeks?

Well, no-one came forward explaining to me how to reach the db state where I can test this, nor the workflow to follow in order to reach the PDFWoPrint.php file by following links within the app rather than typing in the url in the browser.

You can understand I have been loath to waste my time exploring the interface trying to figure that out...

I have now done the following:
- installed a fresh db, with demo data
- went to /SelectWorkOrder.php
- picked 'Melbourne', then search
- clicked on the 'paperwork' link for one item listed
- clicked on both buttons displayed: Label and Paperworks
- checked the generated PDFs - both have 3 qrcodes

Is that what you wanted?

timschofield:
> > This change is still waiting for testing after 3 weeks?
> 
> Well, no-one came forward explaining to me how to reach the db state where I can test this, nor the workflow to follow in order to reach the PDFWoPrint.php file by following links within the app rather than typing in the url in the browser.
> 
> You can understand I have been loath to waste my time exploring the interface trying to figure that out...
> 

Well I have trained hundreds of people to use webERP over the years, and even when the interface isn't in their first language I have found people have had no trouble quickly finding the option they need.

I think I probably felt that a man of your immense talent was at least the equal of those people.

Obviously my bad.

Tim

gggeek:
I know, right? I mean, I like a bit of light hearted forum trolling as much as the next guy, but then sometimes I wonder, am I wasting my time contributing to a project in such a state of quasi-perfection? Surely there must be many more needful projects out there... 😝

timschofield:
Well I think the problem is that we have different views on what constitutes perfection.

Our aim has always been to produce a secure feature rich ERP using simple and easily read code. By and large we have succeeded in doing this. Like any human endeavour, sometimes we fall short of the ideals we set ourselves, but when we find those cases we do our best to improve them.

This guiding principle will not change while I am around. Your contributions are most appreciated when they do not contravene this principle, and this has been true of most of your contributions and I have merged them. On the few occasions that I have judged they do contravene this principle (and somebody has to make that judgement) I have refused the contribution. This is true of most if not all open source projects.

I accept that sometimes I allow my frustrations to boil over, and I may use language that I probably shouldn't, but then I am probably too old to change :)

Tim

PR #676, opened 2025-08-30
made uselogin independent of login web form; ...

gggeek:
...formatting in api_login.php


includes/UserLogin.php
```diff
@@ -275,9 +279,10 @@ function userLogin($Name, $Password, $SysAdminEmail = '') {
 										$EmailText);
 				}
 
-				unset($_POST['UserNameEntryField']);
-				unset($_POST['Password']);
-				unset($_SESSION['DatabaseName']);
+				// gg: moved this into session.php, to make this code independent of the login web form
```
timschofield, line 282: Hi Gaetano, I'm sure I'm missing something but what do we gain from moving this code to session.php?

Tim

timschofield: @gggeek I'm not that bothered about the changes in this PR, though it does seem to be trying to fix something that isn't broken. 
I am more concerned about the direction you are talking about taking this. I am horrified about the idea of users writing command line scripts to update the base accounting records, bypassing the controls that webERP has in place to maintain the integrity of those records. I can't off hand think of one legitimate reason for a user to be doing this. I can however think of many illegitimate reasons!
I strongly believe that right now we need to be concentrating on fixing things in the code that do not work so that we can get v5 out of the door, and then we can look to the future.

Tim

gggeek: @timschofield I think there is a misunderstanding at play.

It is always possible for someone with cli access to the server to run sql scripts, or even type away in a mysql client session, and modify db data, bypassing any security check from the application and all audit logging done from the application.

At the moment, the application does make it hard to create php cli scripts which do carry out long-running or periodic tasks, because many things, but user login in particular, are tied up to the session and to the login web page.

Imho this does actually encourage developer-users to hack around the system and do direct data modifications in their own scripts.

What I am proposing is not to make it easier for developer-users to develop their own hacks, but actually the opposite:
1. make it easier to develop php cli scripts which use the standard login, audit facilities
2. provide php cli scripts for all tasks which are commonly run as cron jobs, instead of relying on a cron-job which accesses a web page via curl (and actually forces disabling auth, as is currently done in at least one case!). Those will need a user/password to be passed in via cli options or an environment variable
3. optionally, take some of the existing "utility" pages, which clearly look like batchy stuff, ie. to be used only once or rarely, and which could take a lot of time and incur in either php timeout or out-of-memory, and transform them into cli scripts (either that, or have cli scripts which carry out the same task as the web page, avoiding code duplication as much as possible of course)

timschofield: > It is always possible for someone with cli access to the server to run sql scripts, or even type away in a mysql client session, and modify db data, bypassing any security check from the application and all audit logging done from the application.

Yes, of course someone with the database username/password combination can do a lot of damage to the system. I don't consider that a reason to give up on trying to protect the integrity of the base accounting records.

> 
> At the moment, the application does make it hard to create php cli scripts which do carry out long-running or periodic tasks, because many things, but user login in particular, are tied up to the session and to the login web page.

Yes, that is by design.

> 
> Imho this does actually encourage developer-users to hack around the system and do direct data modifications in their own scripts.

Yes, it is Free software and users are encouraged to modify existing scripts or make their own scripts, to be run through the webERP interface which provides a structured and controlled access to the base records.

> 
> What I am proposing is not to make it easier for developer-users to develop their own hacks, but actually the opposite:
> 
>     1. make it easier to develop php cli scripts which use the standard login, audit facilities

But what for? Apart from report_runner (which it is well known that I hate and would remove if I could) and the admin scripts that used to be part of the new release process but can now be removed, I can't think of a legitimate need to do this.

> 
>     2. provide php cli scripts for all tasks which are commonly run as cron jobs, instead of relying on a cron-job which accesses a web page via curl (and actually forces disabling auth, as is currently done in at least one case!). Those will need a user/password to be passed in via cli options or an environment variable

I have setup more implementations of webERP accross the world than I can count, and have never set up any cron jobs related to the running of webERP. If you are meaning for the creation of backups then I believe that is the job of the sysadmin.

> 
>     3. optionally, take some of the existing "utility" pages, which clearly look like batchy stuff, ie. to be used only once or rarely, and which could take a lot of time and incur in either php timeout or out-of-memory, and transform them into cli scripts (either that, or add cli scripts which carry out the same task as the web page, avoiding code duplication as much as possible of course)

As I have said several times, running any of the Z_ scripts should be left to somebody who is an expert in webERP. I always disable access to the Utilities module for all the sites I manage. They can do immense harm to the integrity of the data if used wrongly. Making them easier for people to use from the command line is the stuff of my nightmares!

Tim

gggeek:
> I don't consider that a reason to give up on trying to protect the integrity of the base accounting records.

I think that if we consider data integrity and security important, we should actively encourage good practices by end users. That could f.e. be a paragraph or a couple of phrases in the installation manual, mentioning things such as: setting up backups, enabling the mysql query log, making sure the webserver logs are also kept for a relevant amount of time, the standard php.ini settings which affect security, the tightest settings related to fs permissions which work with webERP, the tightest .htaccess which can be use, etc...

> Yes, it is Free software and users are encouraged to modify existing scripts or make their own scripts, to be run through the webERP interface which provides a structured and controlled access to the base records.

I fail to see how making your own scripts to be run through the webERP GUI differs from making your own scripts which are run on the cli, if both go through the same auth layer data access layer and auditing layer.

> I have setup more implementations of webERP accross the world than I can count, and have never set up any cron jobs related to the running of webERP. If you are meaning for the creation of backups then I believe that is the job of the sysadmin.

Just by a cursory grep for `$AllowAnyone`, I found the following web pages which actively encourage users to call them via curl or equivalent, i.e. automate their execution (automation might be of course be triggered periodically or on-demand): MailSalesReport, MailSalesReport_csv, MaintenanceReminders. Possibly also MailInventoryValuation, PDFWeeklyOrders and RecurringsalesOrderProcess.

As it stands, __all of those web pages are a security liability__, as they in fact disable all authorization checking.

Not only that, but because of the twisted way the DB to use is tangled with the session/cookie/login process, there are traces of a harcoded database name (`$DatabaseName`) in pages which disable auth checking :-(

I think that is enough proof both that _some users_ of webERP did find it useful to set up cron jobs, and that the current system makes that hard, brittle and unsafe.

Wouldn't it be better if it was possible instead to set up cron jobs that do not use curl, do proper authorization and authentication, and achieve the same results?

On a personal note: I have never seen a single app on which i worked, as small as it might be, that did _not_ benefit from introduction of cron jobs - the opposite of your experience.
Common tasks which are good to run on a schedule include: culling temporary data; running reports; checking for data integrity (any rules more complex than what the db enforces); sending notifications.

Speaking of temporary data: I saw a few scripts which drop a table, then create it and fill, on every access. I am not sure if that data is meant to stay in the db after those pages finish execution, but it seem that it does (those are not temp tables, and they show up in the demo sql). Is that by design? It seems an unusual pattern...

> As I have said several times, running any of the Z_ scripts should be left to somebody who is an expert in webERP. I always disable access to the Utilities module for all the sites I manage. They can do immense harm to the integrity of the data if used wrongly. Making them easier for people to use from the command line is the stuff of my nightmares!

I am at a loss here.

Some of the Z_ scripts are currently non functional, but present in the system. There is no simple way to figure out which one is broken and which one is not. In fact, quite a few of them seem to have been written for a specific customer, and never made generic enough to be useful for everyone...
It is possible to make them all executable via the web interface, even though, as you said, some are dangerous.
And yet, somehow, making the useable via the command line, while still requiring username/password authentication, is more dangerous???

gggeek:
@timschofield pps: even if you are against command line scripts, there are still the following benefits to this PR:

1. make it easier to introduce non-web-form-auth for APIs in the future
2. reduce lines of code
3. reduce the creep of global state into functions and increase code locality

Those might be considered small or nonexistent. Imho they outweigh the risk of breackage introduced.

timschofield:
> I think that if we consider data integrity and security important, we should actively encourage good practices by end users. That could f.e. be a paragraph or a couple of phrases in the installation manual, mentioning things such as: setting up backups, enabling the mysql query log, making sure the webserver logs are also kept for a relevant amount of time, the standard php.ini settings which affect security, the tightest settings related to fs permissions which work with webERP, the tightest .htaccess which can be use, etc...

I think you are missing the point. It's the accounting integrity I am talking about. There are certain base accounting records that should never be messed with except in a proper controlled manner, which currently the webERP interface provides.

> 
> I fail to see how making your own scripts to be run through the webERP GUI differs from making your own scripts which are run on the cli, if both go through the same auth layer data access layer and auditing layer.

I fail to see why it is necessary to provide command line scripts to update the DB. If I was still auditing and I had a customer who had an ERP system that allowed this, my first recommendation would be that they must find a new ERP system.


> 
> Just by a cursory grep for `$AllowAnyone`, I found the following web pages which actively encourage users to call them via curl or equivalent, i.e. automate their execution (automation might be of course be triggered periodically or on-demand): MailSalesReport, MailSalesReport_csv, MaintenanceReminders. Possibly also MailInventoryValuation, PDFWeeklyOrders and RecurringsalesOrderProcess.

A cursory look through the mailing lists and forums would show that I have always oppose0 the use of $AllowAnyone but it was already being used by many users. However it is only being used to produce reports as your list proves, not to update the system.

> 
> As it stands, **all of those web pages are a security liability**, as they in fact disable all authorization checking.
> 

Agreed, I always opposed the use of $AllowAnyone

> Not only that, but because of the twisted way the DB to use is tangled with the session/cookie/login process, there are traces of a harcoded database name (`$DatabaseName`) in pages which disable auth checking :-(
> 
> I think that is enough proof both that _some users_ of webERP did find it useful to set up cron jobs, and that the current system makes that hard, brittle and unsafe.

Agreed, we should stop it, but they are used by a lot of users.

> 
> Wouldn't it be better if it was possible instead to set up cron jobs that do not use curl, do proper authorization and authentication, and achieve the same results?

Isn't it way way better to stop their use at all?

> 
> On a personal note: I have never seen a single app on which i worked, as small as it might be, that did _not_ benefit from introduction of cron jobs - the opposite of your experience. Common tasks which are good to run on a schedule include: culling temporary data; running reports; checking for data integrity (any rules more complex than what the db enforces); sending notifications.

How many accounting/ERP applications are we talking about that you have worked on?

> 
> Speaking of temporary data: I saw a few scripts which drop a table, then create it and fill, on every access. I am not sure if that data is meant to stay in the db after those pages finish execution, but it seem that it does (those are not temp tables, and they show up in the demo sql). Is that by design? It seems an unusual pattern...

I'm not sure what your experience of MRP functionality is? The MRP functionality generates tables of predicted demand and supply each time it is run. The previous runs are deleted at the start of each new run. This is not an unusual pattern for MRP

> Some of the Z_ scripts are currently non functional, but present in the system. There is no simple way to figure out which one is broken and which one is not. In fact, quite a few of them seem to have been written for a specific customer, and never made generic enough to be useful for everyone... It is possible to make them all executable via the web interface, even though, as you said, some are dangerous. And yet, somehow, making the useable via the command line, while still requiring username/password authentication, is more dangerous???

Using them from the interface is OK as long as it is run under the instruction of someone who has a detailed understanding of webERP. I think I have explained this several times recently...

If all this is so great why don't people like SAP and Oracle allow customers to write random cli scripts to mess with their data?

Tim

gggeek:
@timschofield let's just agree to disagree on this topic - it seems we will never reach an agreement.

pakricard:
> and have never set up any cron jobs related to the running of webERP.

I run multiple cronjobs daily, directly related to webERP. They clean, speed up, organize and prepare work for the next day. I consider them an essential tool.

As examples:

- set reorderlevel to 0 to items marked during the day as obsolete
- check users have not messed up and set reorder levels for certain locations
- removed items marked as obsolete from the sales categories tables
- update reorder level values for all current items 
- delete empty internal requests
- automatically authorize internal request generated during the day
- block inactive users not logging in for more than X days
- purge several tables
- prepare and send by email all inventory transfers automatically 
- send emails of tasks to be done to staff
- sync webERP DB with OpenCart DB
- ... 

> Z_ scripts should be left to somebody who is an expert in webERP. I always disable access to the Utilities module for all the sites I manage. They can do immense harm to the integrity of the data if used wrongly. Making them easier for people to use from the command line is the stuff of my nightmares!

I simply do not upload any of these Z_ scripts (except the ones that changes item codes, location codes) as they are extremely dangerous. I strongly believe they should be deleted. In a normal world, you should not need them. 

Regarding accounting integrity: 
Someone with the DB credentials and willing to do harm won't care if they have or not command line or cron job or script access. They can directly go to phpMyAdmin or CLI and update/delete whatever in gltrans and accounting integrity gets compromised. 

I consider hardcode the DB name in some scripts (and give them security page 15 aka sysadmin) very dangerous. The proposed changes in this PR do not decrease security, just sets the first steps towards increased security.

my 2 cents
R

dalers:
(edit, forgot I was forced to use SAP at Baker Hughes)

This is mostly a rebuttal to Tim's points (with respect) as I generally agree with @pakricard (and @gggeek with some cavaets).

> There are certain base accounting records that should never be messed with except in a proper controlled manner, which currently the webERP interface provides.

I would suggest then that NO files be provided with webERP other than files which can be selected directly from the user interface. Also related, any files which are known to have issues or which could compromise the integrity of the accounting records should NOT be included in the webERP project ("Z scripts", bash scripts and the like could still be provided in a seperate companion project).

> I have always oppose0 the use of $AllowAnyone

@timschofield I don't need to remind you that you are the project lead and personally control the project repo. If you don't believe webERP should support $AllowAnyone then you have the ability and ethical responsibility to remove the capability from the project. However, if you believe the needs of the webERP community outweigh your personal opinion in this regard, then we should continue to provide the feature but implemented in a responsible secure manner.

> as long as it is run under the instruction of someone who has a detailed understanding of webERP

Imho this is impossible practically. If we are being responsible, we should either not provide such files at all or provide them with clear instructions/warnings, user authentication and logging in the audit trail.

> why don't people like SAP and Oracle allow customers to write random cli scripts to mess with their data?

This is a red herring. The real reason is related to the complexity of their software, service contracts and the goal of commercial software vendors to maximise profit, minimise liability and protect a closed eco system of service providers. Also no responsible corporate officer (financial or otherwise) would entertain having employees write and run "random cli scripts" .

Fwiw, I have used SAP (but not Oracle) and several commercial 2nd-tier systems (Syspro, Syteline, Infor Visual and Dynamics GP) and knowlege through colleagues of a few more (Dynamics NAV, Epicor, Exact, Made2Manage and MRPEasy). I have also participated in two implementation/migration projects as the engineering domain specialist (Syteline and the other I've forgotten). My experience is that a commercial 1st and 2nd tier ERP software implementation generally involves a dozen contractors (a PM, several BIs and a half-dozen developers, each a narrow specialist), in addition to the dozen or so internal domain specialists, which puts them in a different league than companys' for which webERP would be of interest.

dalers:
> I simply do not upload any of these Z_ scripts (except the ones that changes item codes, location codes) as they are extremely dangerous. I strongly believe they should be deleted. In a normal world, you should not need them.

This reminds of of the MariaDb/MySQL "mysql_secure_installation" script. I recommended in my previous comment that "dangerous" files should simply be deleted from webERP but perhaps a "weberp_secure_installation" script to use after installation would be a more palatable approach. 

> my 2 cents

@pakricard you are far too modest. So far as I can tell, you are the _only_ active project contributor personally using webERP to manage the operations of your company. Imho, that's worth $2M, not 2 cents!  :-)

gggeek:
> perhaps a "weberp_secure_installation" script to use after installation would be a more palatable approach.

Which, ironically, would have to be a cli script, as it would need to be run as root and not as the webserver user, in order to remove write access to the `companies` folder, `config.php` and co. 🤣 

I'm all in favour of that, but tbh I'd rather start with creating a clear guide about securing weberp, and showing it (or a link to it) as last step of the setup wizard.

The reason is that there are many aspects of "security" which can not be achieved by an automated script, such as eg. setting up periodic mysql backups. And for things which could realistically be achieved, such as writing an `.htaccess` file to the webRRP root dir, I doubt that we would ever reach consensus about what is the correct/safest approach (it seems that we start from very different positions regarding what are to be considered threat scenarios).

dalers:
> from very different positions regarding what are to be considered threat scenarios.

How would you summarize our positions? 

I think the significant threat scenarios are:

- incorrect usage by authorized users who may be uninformed and perhaps careless but not malicious
- code errors
- unauthorized access from the internet

webERP server sysadmins are not considered threats.

timschofield:
> @timschofield let's just agree to disagree on this topic - it seems we will never reach an agreement.

OK, I was tired and frustrated last night, and I should have used more temperate language - sorry. In my defense I had just spent ages trying to work out why after your changes the installer had suddenly started to output hundreds of messages all saying that the admin user had been created (though the problem turned out to be nothing to do with the admin user). It was obvious you hadn't run the installer before submitting you changes and this meant that my "webERP time" had been spent tracking down why this had happened and then putting a temporary fix in just to make the installer at least function. I was frustrated. I do think that if we are to maintain the integrity of webERP we should as a bare minimum run the scripts we have changed before issuing a PR.

gggeek:
> ... if we are to maintain the integrity of webERP we should as a bare minimum run the scripts we have changed before issuing a PR.

I agree. I am sorry about the breackage. It was very obviously a copy-paste snafu, after I looked at your fix.

I admit that I have not been careful enough in testing my changes so far.

In my defence, running the installer takes quite some time.

This is why I have been steadfastly trying to implement automated testing for it, so that I can avoid inadvertently breaking it.

gggeek:
> How would you summarize our positions?

I can not really voice for any positions except mine. I can try, but please forgive me if I am far off.

> I think the significant threat scenarios are:
> * incorrect usage by authorized users who may be uninformed and perhaps careless but not malicious
> * code errors
> * unauthorized access from the internet
> * webERP server sysadmins are not considered threats.

I agree with the above.

Further to that, my position is regarding cli scripts is:
1. we should agree on whether we allow developer-users to develop their own custom code (be it cli scripts or custom web pages) and/or modify the webERP code
2. if we support / accept that developer-users might introduce such tweaks, we should give them tools to make it easy to develop their own code in a way that insures it goes through at least access-control (aka. login) and controlled data-access apis (which include audit logging)
3. if we do not consider that developer-users tweaking the code is officially supported, then the only reason to make it easy to develop cli scripts is if we want to add to webERP officially sanctioned cron jobs / batch scripts
4. since the webERP code is flexible enough to assign policy checks to any web-page found on disk (not 100% sure about this, but I think it is), including custom ones, it stands to reason that the same principle could be applied to any php cli script found on disk as well
5. there is no difference whatsoever in securing a php cli script vs. securing a web page - in both cases, either the php code starts off with inclusion of session.php and complies with the webERP apis, or it is free to do anything with data and leave very little trace of its actions
6. further to the above, there is currently hackish logic in webERP which was added in order to allow automated execution of some reports which basically bypasses authentication+authorization. In order to undo that, and make code safer, the best solution is to have said reports generated a) via cli scripts, and b) have those cli scripts use the standard auth logic

Finally, back to the threat models:
* while "webERP server sysadmins are not considered threats", we should strive to a) make it hard for such sysadmins to do the wrong things, easy to do the right things, b) use as much as possible the facilities provided by the OS (such as filesystem permissions) to harden and secure the app from unintentional mistakes and side attacks (eg. another user of the operating system who has nothing to do with webERP should never be able to have any effect on the system)
* atm there is, imho, little care given to the "malicious access from the internet" case. The code is not developed with a defensive mindset, but it rather assumes that the http requests it receives are only the ones generated by clicking on links/forms within the app
* re "code errors": the "defence in depth" principle always applies. A php site can be hacked because of errors in the php code, in the php dependencies, in the php engine, in the myriad of C libraries loaded by the php engine, in the web server code, in the db code, in the OS code. Of those, only the first one is under control of the webERP developers. Measures should be taken to mitigate as much as possible breaches in all the others (fe. we could recommend that the sysadmins set up logging of all commands executed by a shell, that the os logs are shipped to an external system via syslog or written to an append-only filesystem, that the mysql server logs 100% of the executed queries. BUT all of those measure take a lot of time, generally make usage of the system more unwieldy, and are hard to set up properly. Given the fact that webERP target usage is the SME, it seems unlikely that such hardening will be often applied).

pakricard:
> 1. we should agree on whether we allow developer-users to develop their own custom code (be it cli scripts or custom web pages) and/or modify the webERP code

By definition of open source, developers-users can and will tweak and develop their own custom code. I am one of those ;-). 
No way to avoid it. Then, the safer the environment, the better. 

> 2\. if we support / accept that developer-users might introduce such tweaks, we should give them tools to make it easy to develop their own code in a way that insures it goes through at least access-control (aka. login) and controlled data-access apis (which include audit logging)

Agree

> 3\. if we do not consider that developer-users tweaking the code is officially supported, then the only reason to make it easy to develop cli scripts is if we want to add to webERP officially sanctioned cron jobs / batch scripts

Officially supported or not, it happens ;-). If we create the proper setup, then we can distribute as well some cron jobs or bash scripts if needed.

timschofield:
I think there is soime confusion here. When I talk about the security and integrity of the base accounting records I mean from an accountancy perspective. Accountants work within a set of globally agreed standards, and where something might be technically possible to do, if it controvenes those standards we can't do it.
One of the things I used to get asked for most when I was installing webERP around Africa was the ability to edit an already posted transaction. Whilst this is easily doable from a coding perspective it destroys the integrity of the base accounting records and is a fraudsters charter. I know we can't prevent a determined user doing it, but making it easier to do via random command line scripts is not something I want webERP to do.

Tim

timschofield:
> @timschofield I don't need to remind you that you are the project lead and personally control the project repo. If you don't believe webERP should support $AllowAnyone then you have the ability and ethical responsibility to remove the capability from the project. 

Oh how I wish life was this simple. I also have a responsibility to not break current implentations if at all possible, and many implentations still use it. Removing $AllowAnyone would break those implentations so I hold my nose and just accept that this is the way the world is :)

gggeek:
> I think there is some confusion here. When I talk about the security and integrity of the base accounting records I mean from an accountancy perspective. Accountants work within a set of globally agreed standards, and where something might be technically possible to do, if it contravenes those standards we can't do it.

I think no-one in this discussion or other similar ones proposed to relax the current security and integrity practices, esp. in ways that would make the app legally unsuitable for usage, whichever the country.

On the other hand, if something is required from a legal / accountancy perspective, we can not abstract from the limits/issues of its technical implementation - those can not be simply waived away, f.e. if something bad has happened and there are court proceedings involved. We should at least develop according to "well established",  "industry standard" security practices.

The setup wizard was f.e., up until to a very recent commit, suggesting a default password at install time. That would, arguably, run afoul of PCI DSS requirement 2.

> [...] I know we can't prevent a determined user doing it, but making it easier to do via random command line scripts is not something I want webERP to do.

I will repeat my opinion on this once more:

* nothing in webERP is, atm, preventing a zealous syadmin, developer or implementor from adding custom code, be it bash scripts, pgp cli scripts or web pages*.
* The main difference between one such user creating a custom web page vs. a custom cli script is that, atm, it is easy to make the web page do authentication and authorisation, while it is hard to make the cli script do the same.
* It seems that multiple users, including @pakricard, have felt the need over time to run periodic tasks. That resulted in disabling access controls for a select list of report-generating pages, which we all agree is a bad solution. A better solution would have been, imho, allowing those reports be generated via _cli scripts that respect access controls, and require a login/password to run_.
* Even if we make it easy to develop cli scripts which respect access controls, nothing prevents the security-minded sysadmin/product-owner to lock down the app by setting filesystem permissions to read-only, restricting shell access, and tightening access to the existing php scripts via the weberp policy system. Those 3 things are to be done (or not required) in any case, irregardless of said ease-of-development.
* Vice-versa, there is a chance that current users _already_ developed cli scripts which modify data, and do so in a way which make it completely untraceable. Which is what we all want to avoid.

I fail to see, in this long thread, where you have confuted, at least once, the fact that cli scripts can be made to respect the same access controls than web pages. Or explain why you think that they are inherently less safe than web pages (when, if anything, they can only be executed only by someone with shell access, which is, generally, a more restricted set of users than the users of the web interface).

I am happy to discuss this further by voice, if you deem it useful, otherwise I guess we will just have to accept that we do not agree on this whole topic.

`* = of course barring blocking write-permissions to the filesystem. But there is no "tighten down the installation" guide available atm, afaik, with steps by step instructions about that. Even if there was one, the code currently does make it hard to set the whole installation to read-only, as there is a lot of scripts which generate pdf files or other reports, and not all of them are stored in the system temp dir...

timschofield:
How about we agree to park this until we get v5 out of the door and then sit down together and work out a way between us to make this work?

gggeek:
> How about we agree to park this until we get v5 out of the door and then sit down together and work out a way between us to make this work?

Sure.

I'd still like to see this one go on quickly, as it is a small change, and does not seem to carry big risks, but it's not a blocker either.

Side note: I hope I'm not distracting anyone from work required for the 5.0 release. On my side, all I am working on that I'd like to see go in before that is: 1. fix the installer unit tests / clarify minimum supported db version; 2. bring back the change of default db tables charset to utf8mb4; 3. add an 'add company' mode to the installer; 4. a couple of installer improvements for robustness and user experience; 5. double-check if the current usage of gettext commands to extract translation strings from php files still works or needs fixing.

timschofield:
>5. double-check if the current usage of gettext commands to extract translation strings from php files still works or needs fixing.

Ah, just realised that it doesn't work. xgettext works by extracting all the strings within the _() function. Your __() function does a great job of translating when the locale isn't present but doesn't work with xgettext.

Problem....

gggeek:
Xgettext takes an arg or cli option telling it which function to look for. I dont remember off the top of my head, but it should be in the docs of the polyfill-gettext package

timschofield:
> Xgettext takes an arg or cli option telling it which function to look for. I dont remember off the top of my head, but it should be in the docs of the polyfill-gettext package

OK, I read the man file earlier and couldn't see anything but you know more about this stuff than me. I'll leave it with you.

PR #682, opened 2025-08-31
add plenty of devs docs

gggeek:
The dev workflow was taken from the website, reformatted by hand. I will send a 2nd pr with the images

The coding standard was taken from the website, reformatted by hand, then
- added a section at the top listing the "surprising" choices
- moved the paragraph about using `prnMsg` to the cookbook doc, which is meant to be more of a hands-on guide to implement common tasks


gggeek:
Added a commit with the missing images, as well as updated to README and to INSTALL

timschofield:
These are mostly good, but you are an intelligent guy, you know I will not allow those changes to the coding standards that attack the project. 

Why not just state what our standards and guidelines are, and anybody smart enough to understand it can work out any differences themselves.

Tim

gggeek:
@tim I did my best not to be snarky, dismissive, insulting or demeaning when writing the section about "things you might find surprising".

I think that there is value in having that section in the docs - it would for sure have saved me a lot of time arguing for the introduction of many changes if I had found out on day one that all those "unexpected" (by me) practices were known and accepted.

I am happy for you to modify this PR and rewrite the docs as you see fit.

timschofield:
Well I just thought that it is obvious just by looking at the code what the coding style we use is, and have been using for 22 years. 

BTW the 4 tests you wrote are all failing since you updated the installer, can you check?

Tim

gggeek:
> Well I just thought that it is obvious just by looking at the code what the coding style we use is, and have been using for 22 years.

Most of the items in the list are obvious, but at least a couple caught me off guard, such as the escaping of data from users being done within session.php, and the fact that different files define different classes with the same name.

Also, the fact that most of those things are obvious does not necessarily mean that they are something that the project considers good practice - they might either be historical accidents and something that it might be welcome to have fixed in the future (such as eg. haphazard formatting), or deliberate choices that it is useless to question (such as eg. the flat folder structure).

As for "we have been using it for 22 years": I don't understand fully your position about this - it feels as if you are at the same time proud of the way things work, and thinking that listing the way things are not supposed to work is an attack on the project.

> BTW the 4 tests you wrote are all failing since you updated the installer, can you check?

I know, sorry about that. I have been working until late night trying to fix them.
I have a code branch which fixes 2 of those, and will look today into the other 2.
In fact, I think that at least 1 of them fails because there is some sql code which makes mysql 5.5 unhappy (the tests are now running on mysql 8.0.43, mysql 5.5, mariadb 5.5 and mariadb latest).
If confirmed, this might be an interesting but if info for the discussion about the minimum mysql version to support.

timschofield:
Pretty sure that nobody would think that "haphazard formatting" is a deliberate project policy. I rather think that they would conclude that it is as a result of an open source project pulling in contributions from many people in the past, not all of whom have formatted the code correctly.

We have always tried to make our code legal PHP, and I think that apart from occasional bugs that get fixed when they are spotted, we succeed in that. Whenever a new version of PHP has come out we have fairly quickly tried to make our code compliant with that version - I think historically we have been quicker on that than most projects. 

Our code is legal PHP and up to date, so I think having a "things Gaetano doesn't like about our coding guidelines" section in the developer docs is unhelpful, and I think you know that.

It's a shame because the rest of the PR is good.

gggeek:
> Pretty sure that nobody would think that "haphazard formatting" is a deliberate project policy

I am not sure I put that phrase in the docs... atm they state `the formatting standards outlined above have been followed very loosely`

Which is a fact, and stated in as neutral a tone as I could conceive.

I also think that they way I wrote those items does not try to apportion blame, nor to pass a judgement.

Everyone who has just a tiny bit of experience in development knows how hard it is to keep a codebase well formatted, esp. when many hands are involved.

As for it being a deliberate policy, I did not think to be suggesting that this is a deliberate policy - even though it is also a fact that there seems to be little appetite for carrying out a reformatting pass or introducing reformat-on-commit tools.

Not that I am a fan of automated formatters, mind you - I prefer to have the freedom to break the rules in 5% of the cases, where custom formatting seems to make the core more readable.

> things Gaetano doesn't like about our coding guidelines

You might be misunderstanding. A lot of the things in the list I am not a fan of, and would not suggest adopting even in a green-field project. I just think that they are the "latest fashion" in php development.

timschofield:
> > Pretty sure that nobody would think that "haphazard formatting" is a deliberate project policy
> 

It is in a previous message on this thread.

> As for it being a deliberate policy, I did not think to be suggesting that this is a deliberate policy - even though it is also a fact that there seems to be little appetite for carrying out a reformatting pass or introducing reformat-on-commit tools.
> 

I tried using php beautifier to do this via a hook (it is still there but disabled in https://github.com/timschofield/hooks) for a while but stopped for various reasons, not least because I prefer to be in charge of changes myself, not having something change the code without my seeing it first.

Tim

dalers:
Looks good except i would want to see the **Deviations from PHP "best practices"** heading removed as "best practices" is subjective and often changes over time. The points can be kept as explanations for appropriate standards. I've only skimmed the source for the docs though and would need to view the rendered markdown to give approval.

However, before there is any value to discussing wording it must be decided if the development guidelines and coding standards are to be moved from the website into the webERP source. Imho this would be an improvement as the benefit of tightly coupling the documentation to the code and the ease of access to a developer outweighs a philosophy of prefering code only in the repo.

gggeek:
> i would want to see the Deviations from PHP "best practices" heading removed

Happy to oblige. Or change that to "latest fashions", which seems to better reflect the consensus of the webERP dev team

> it must be decided if the development guidelines and coding standards are to be moved from the website into the webERP source

see the discussion ongoing on the open PR in the website github...

dalers:
> Or change that to "latest fashions",

Just as subjective. Imho coding standards should be statements of fact with as little room for interpretation as possible.

P.S. PR to weberp.org that @gggeek referred to: https://github.com/timschofield/weberp.org/pull/12

gggeek:
In the spirit of constructive collaboration, and to allow advancing on this topic, I have now removed the controversial paragraph.

I hope this can now go in.

Issue #450, opened 2025-06-22
issues with the installer


gggeek:
A few issues I found while trying to set up the application:

- no error messages are displayed at all if wizard pages fail halfway through because of a php error. While that is good for safety, we should try to at least mention to the user than an error happened
- if the database tables already exists, they are not dropped, and the wizard errors out. It would be nice to have an option (with confirmation) allowing to drop them
- the wizard does not check if the db user account can create triggers, which is often not the case in mysql. When that happens, the php error log contains a message similar to "You do not have the SUPER privilege and binary logging is enabled (you *might* want to use the less safe log_bin_trust_function_creators variable)"
- if the config.php file exists, and the user navigates to the setup wizard pages, the setup process is started again, without any warning that the config was already create previously

dalers:
Thanks for your impressions. @timschofield designed and implemented the "new" install system for webERP v5 and wrote an [overview of the installer in the wiki](https://github.com/timschofield/webERP/wiki/Installer) 

More generally, what do you think of the strategy of using individual SQL files for each table schema and also for default table data?

timschofield:
Hi Gaetano, as @dalers says, this is a complete rewrite of the installer so I am glad you are testing it. My answers are written inline below.

> A few issues I found while trying to set up the application:
> 
>     * no error messages are displayed at all if wizard pages fail halfway through because of a php error. While that is good for safety, we should try to at least mention to the user than an error happened

Have you specific examples of this?

> 
>     * if the database tables already exists, they are not dropped, and the wizard errors out. It would be nice to have an option (with confirmation) allowing to drop them

This was something I thought about a lot when writing the new installer. I didn't like the idea that existing tables and their data could be accidentally or deliberately dropped. Other than for testing purposes it was hard to envisage a situation where a user would want to legitimately install over an existing database, so I left it to error out, if it was tried. Am happy to hear alternative views on this.

> 
>     * the wizard does not check if the db user account can create triggers, which is often not the case in mysql. When that happens, the php error log contains a message similar to "You do not have the SUPER privilege and binary logging is enabled (you _might_ want to use the less safe log_bin_trust_function_creators variable)"

You are correct on this. @pakricard  introduced the triggers after I rewrote the installer and I never went back and looked at this. I guess we should check for this at the point we are checking the DB connection. Do you know off hand the code to check this?

> 
>     * if the config.php file exists, and the user navigates to the setup wizard pages, the setup process is started again, without any warning that the config was already create previously

[...]

PR #683, opened 2025-08-31
installer: fix checks for existing tables an abort if any are found; ...

gggeek:
...prevent overwriting config.php; remove flawed checks for permissions to create a schema; minor other fixes


gggeek:
Fixes 3 of the remaining 5 tasks for ticket #450

Issue #687, opened 2025-09-01
Issues trying to run the installer on mariadb 5.5


gggeek:
Afaik there is no 100% clear consensus on the minimum supported versions of mysql and mariadb.

I have started running tests on mysql 5.5 and 8.0.43, mariadb 5.5 and latest.

Atm all of them fail (using the installer from `master`) :-(

For mariadb 5.5 at least, the issue comes from unsupported SQL.

These are the errors:

```
table debtortrans

Warning: mysqli_query(): (42000/1064): You have an error in your SQL syntax; check the manual that corresponds to your MariaDB server version for the right syntax to use near 'STORED, PRIMARY KEY (`id`), KEY `DebtorNo` (`debtorno`,`branchcode`), KEY '

table loctransfers

Warning: mysqli_query(): (42000/1064): You have an error in your SQL syntax; check the manual that corresponds to your MariaDB server version for the right syntax to use near 'STORED, KEY `Reference` (`reference`,`stockid`), KEY `ShipLoc` (`shiploc`), '

table salesorderdetails

Warning: mysqli_query(): (42000/1064): You have an error in your SQL syntax; check the manual that corresponds to your MariaDB server version for the right syntax to use near 'STORED, PRIMARY KEY (`orderlineno`,`orderno`), KEY `OrderNo` (`orderno`), '

table sessions

arning: mysqli_query(): (42000/1064): You have an error in your SQL syntax; check the manual that corresponds to your MariaDB server version for the right syntax to use near '&quot;&quot;, `scripttime` TIMESTAMP NULL, PRIMARY KEY (`sessionid`) ) ENGINE=InnoDB D' 

table stockmaster

Warning: mysqli_query(): (42000/1064): You have an error in your SQL syntax; check the manual that corresponds to your MariaDB server version for the right syntax to use near 'STORED, PRIMARY KEY (`stockid`), KEY `CategoryID` (`categoryid`), KEY `Des'
```

gggeek:
PS: in accord with recent discussions about SQL MODE to use, I am now using this SQL MODE to run the installer when doing automated tests:

`STRICT_ALL_TABLES,ANSI,ERROR_FOR_DIVISION_BY_ZERO,NO_AUTO_VALUE_ON_ZERO,NO_ENGINE_SUBSTITUTION,NO_ZERO_DATE,NO_ZERO_IN_DATE,PAD_CHAR_TO_FULL_LENGTH`

dalers:
[...]

Issue #692, opened 2025-09-02
PR #688 commit caused "Continuous Integration: All jobs have failed" on my branch


pakricard:
Hi:

I received an email from GitHub a couple of hours ago stating:

<img width="550" height="669" alt="Image" src="https://github.com/user-attachments/assets/6342a422-07c6-4f1e-bb05-c2ef451123cc" />

Sniffing around, it appears that commit https://github.com/timschofield/webERP/commit/01bec890d4ede5252c68243081322313709efdae for PR #688 caused it.

I have absolutely no clue what is it about ;-)

R

gggeek:
There's been a lot of evolution, concurrently, of the automated tests (which so far only test the installer) and the installer.

The tests now run on 4 different db versions. And they fail on mariadb 5.5 and mysql 5.5 (see other ticket).

If your branch is up to date with master, the tests running on php 8.4 and 8.3 should pass... did you merge recently?

pakricard:
> did you merge recently?

I received the email with the errors after I merged Tim's branch yesterday.

Today, I pulled and merged again and only shows errors related to MariaDB 5.5 and mySQL 5.5 which I think is consistent with the issues you try to fix regarding the installer.

<img width="509" height="601" alt="Image" src="https://github.com/user-attachments/assets/a12cccad-5c95-4933-a4d4-214748ec0dd7" />

R

pakricard:
I still get errors when committing:

<img width="641" height="420" alt="Image" src="https://github.com/user-attachments/assets/56d14c93-c924-4d65-8e24-c5dac470d7f1" />

gggeek:
Yes.

If you look in detail, you will see that now only one test case is failing: the one on mysql 5.5.

Yesterday s nerge fixed the mariadb 5.5 one.

Please bear patience for a little while more

pakricard:
> Please bear patience for a little while more

Sure, no worries! It was just a "FYI"

R

gggeek:
Habemus papam?

<img width="1960" height="852" alt="Image" src="https://github.com/user-attachments/assets/935ea246-38b0-4436-9bd6-cd557d4d0668" />

gggeek:
@pakricard can this be closed?

pakricard:
> Habemus papam?

Habemus! issue can be closed.

pakricard:
@gggeek : Somehow, when I merged the PR #665 it caused again the same warnings, even in Tim's branch. 
[...]

PR #698, opened 2025-09-04
bump min mysql version to 5.7.5: update tests and docs; ...

gggeek:
...fix a few places in code where db type mariadb was forgotten


gggeek:
Should fix issues #687 and #692

PR #714, opened 2025-09-17
fix Login.php warnings when included from non-top-level pages

gggeek:


timschofield:
I don't mind merging this, but I am curious to know what it fixes as I see no warnings.

gggeek:
I got the warnings when going directly (ie. typing their url in the browser) to dashboard and reportwriter pages as anon user. I suspect that it also broke the listing of companies in the login form for those cases.

timschofield:
I am confused, what does "as anon user" mean?

gggeek:
"as anon user" = not having logged in

timschofield:
If you haven't logged in and you try to go to Dashboard.php then it should take you straight back to the login page:
https://www.weberp.org/demo/Dashboard.php

Are you saying on your system it doesn't?

gggeek:
The URL I am talking about is not `/Dashboard.php`, it is f.e. `/dashboard/bank_trans.php`.

There is code in all the dashboard scripts which is meant to allow them to be run either as "applets" or as standalone web pages. Not that I think that those make necessarily a lot of sense as standalone pages, but I think that in case anyone tries to access them as anon user, the login form should be displayed correctly.

While at it, I think t might be worth discussing a bit more what happens when users try direct access to _any_ of the php files of webERP which are not meant to be standalone web pages, eg. all files in `/include`, all files in `/api` bar one, `/config.php` etc...

In my opinion:
1. none of those should display a php error/warning (and atm a lot of them do)
2. ideally, they should all respond with either a 404 or a 302 answer

Given we already discussed adding .htaccess file in the sources and you are against it, I think the best option to implement point 2 above is to add a bit of code at the top of (almost) every such php script.
Also, if we implement point 2, we fix point 1 in the course of action.

I know it's a lot of files to touch, but I think it's worth it.

PR #715 is part of that.

gggeek:
ps: the issue with accessing directly a non-top-level-url as anon user and Login.php getting things wrong is not only affecting dashboard applets - it also affects urls such as `/reportwriter/ReportMaker.php`.

In fact, it does affect every php script which is not in the top-level dir.

This fix is just one more part of the long-running task "fix usage of $PathPrefix and $RootPath".

timschofield:
Well including a .htaccess file doesn't help the increasingly large number of people using nginx.

Isn't it the job of any self respecting sysadmin to just stop the users accessing any files they shouldn't? I know I always do when setting up a system, but if someone is stupid enough to try to access one of these scripts without logging in, then all that happens is they get some error messages they won't understand??

gggeek:
> Well including a .htaccess file doesn't help the increasingly large number of people using nginx.

No, but it does serve as a blueprint for people using other webservers. Atm webERP does not provide a clear, concise list of which folder should or could be forbidden from web requests.

> Isn't it the job of any self respecting sysadmin to just stop the users accessing any files they shouldn't?

How are they supposed to know which ones? webERP does not separate them neatly into a "public" folder and "everything else". (It also does not say which folders should be writeable for reports, or for Z_ scripts, btw, but let's not open that other can of worms now)

>  but if someone is stupid enough to try to access one of these scripts without logging in, then all that happens is they get some error messages they won't understand??

It's not about stupid users. It's about robust and secure software. Getting detailed error messages is considered something valuable for attackers. Which might be hackers from over the internet, or even disgruntled employees when the app is locked down to an intranet.
We could add some docs stating "always set display_errors to false after install" but a) users seldom read docs, and b) why force users to jump through hoops when we can build things which are safe out of the box?

ps: I don't think this discussion bears in fact much weight on the fact that this PR fixes code which is evidently buggy (as in: not working correctly for any web page which is not in the web root, of which there are a few).

gggeek:
Re "Isn't it the job of any self respecting sysadmin ...": isn't webERP target of SMEs and coding-accountants a prime group of users who are by definition not "self respecting sysadmins", and who might have troubles finding (and paying) any such professionals - the same way they seem to have trouble finding self respecting coders willing to improve the code on their behalf?

timschofield:
> No, but it does serve as a blueprint for people using other webservers. Atm webERP does not provide a clear, concise list of which folder should or could be forbidden from web requests.

So the software wouldn't meet your criteria of "robust and secure" for nginx users. This is the problem when you start trying to hand hold users to the nth degree.

> 
> How are they supposed to know which ones? webERP does not separate them neatly into a "public" folder and "everything else". (It also does not say which folders should be writeable for reports, or for Z_ scripts, btw, but let's not open that other can of worms now)

That's the beauty of our file structure, it's obvious which ones.

> It's not about stupid users. It's about robust and secure software. Getting detailed error messages is considered something valuable for attackers. Which might be hackers from over the internet, or even disgruntled employees when the app is locked down to an intranet. We could add some docs stating "always set display_errors to false after install" but a) users seldom read docs, and b) why force users to jump through hoops when we can build things which are safe out of the box?

I'm not personally aware of anybody who allows unrestricted access to their webERP over the internet. That would definitely be stupid!!

Can you desctibe what info the error messages could give that the hackers couldn't get from simply going to github and reading the code?

> 
> ps: I don't think this discussion bears in fact much weight on the fact that this PR fixes code which is evidently buggy (as in: not working correctly for any web page which is not in the web root, of which there are a few).

Actually I said that at the start of the conversation. The PR doesn't do any harm - but I can't see how it does any good either, but I am prepared to merge it anyway.

gggeek:
Thanks for merging.

Re. the rest:

> So the software wouldn't meet your criteria of "robust and secure" for nginx users. This is the problem when you start trying to hand hold users to the nth degree.

Indeed my 1st choice would be to move web pages into a `/public` folder, but that is a tremendous amount of work and not likely to be welcome even if I was to start working on it...

Also, I do not think that "making software more secure oob for a part of users, and easier to secure for the rest" is something to scoff at.

> That's the beauty of our file structure, it's obvious which ones.

I strongly dissent.

> I'm not personally aware of anybody who allows unrestricted access to their webERP over the internet. That would definitely be stupid!!

Then you should probably update the installation and requirements docs stating that webERP should never be run on a public network unless an extra layer of access control is added on top.

> Can you describe what info the error messages could give that the hackers couldn't get from simply going to github and reading the code?

It's in the OWASP top-10, currently at position 5 - see: https://owasp.org/Top10/A05_2021-Security_Misconfiguration/ "Error handling reveals stack traces or other overly informative error messages to users".

Talk to any professional php developer, or take a look at any other shrinkwrap php application - they will all agree that production setups should _not_ display php errors to end users.

timschofield:
> Re "Isn't it the job of any self respecting sysadmin ...": isn't webERP target of SMEs and coding-accountants a prime group of users who are by definition not "self respecting sysadmins", and who might have troubles finding (and paying) any such professionals - the same way they seem to have trouble finding self respecting coders willing to improve the code on their behalf?

Now you are being silly @gggeek , and I'm not sure whether it's intentional or not.

The target group is people who are experts at ERP (frequently but not exclusively accountants) in all sorts of organisations (I've never personally said anything about SMEs) to take the code and modify it for their own purposes. That doesn't mean they can't afford to employ self respecting coders who know nothing about ERP to try and modify the software for them, just that they can if they want do that. We have had great success in getting such people to contribute, because we have a coding style that makes the code easy for them to read and understand. I know you don't like that style but history is on my side on this.

You are a way better PHP coder than I am, but when it comes to ERP knowledge and experience I suspect I win out.

timschofield:
> It's in the OWASP top-10, currently at position 5 - see: [owasp.org/Top10/A05_2021-Security_Misconfiguration](https://owasp.org/Top10/A05_2021-Security_Misconfiguration/) "Error handling reveals stack traces or other overly informative error messages to users".
> 
> Talk to any professional php developer, or take a look at any other shrinkwrap php application - they will all agree that production setups should _not_ display php errors to end users.

But can you let me know what info can be gleaned from the error messages that cannot be found more easily by just visiting GitHub and looking at the code?

gggeek:
> But can you let me know what info can be gleaned from the error messages that cannot be found more easily by just visiting GitHub and looking at the code?

For one, the exact folder of installation. It can help in creating scripts which f.e. drop payloads to disk which can later be accessed from the web, or a myriad of other things

timschofield:
> > But can you let me know what info can be gleaned from the error messages that cannot be found more easily by just visiting GitHub and looking at the code?
> 
> For one, the exact folder of installation. It helps in creating scripts which f.e. drop payloads to disk which can later be accessed from the web, or a myriad of other things

Well I would suggest if you have given a hacker access to your web server so they can install scripts on it, they have probably already worked out such things! You have suggested a good reason for not allowing command line scripts near a webERP installation though ;)

gggeek:
> Well I would suggest if you have given a hacker access to your web server so they can install scripts on it, they have probably already worked out such things! You have suggested a good reason for not allowing command line scripts near a webERP installation though ;)

May I suggest you get an education in secure web coding and practices before making such statements?

timschofield:
> > Well I would suggest if you have given a hacker access to your web server so they can install scripts on it, they have probably already worked out such things! You have suggested a good reason for not allowing command line scripts near a webERP installation though ;)
> 
> May I suggest you get an education in secure web coding and practices before making such statements?

Really?? So if I had this education I would then believe that it would be OK to give hackers access to a web server.

OK, I bow to your knowledge :D

gggeek:
@timschofield now it's you getting silly...

For the TLDR:

- php cli scripts vs. php web pages have nothing to do with this
- in the case of webERP, by default, there is a huge unprotected surface - all php scripts are accessible to anyone who can send an http request to the app
- you seem strongly against making it easier for the sys-admin to lock down the system via reducing the number of php scripts accessible via http requests. I still have a hard time figuring out why
- there is no documentation in the project, as far as I can tell, which strongly encourages users to set up 2ndary auth means after install, or to not allow access to the server from the internet

This situation is why it seems to me that the current and previous crop of webERP developers were not security-minded.

As a reminder, it took me less than 2 days to find 2 security-related issues within the v5 codebase, just by looking at the code (and I am not sure those were patched btw...)

As for the specifics of php warnings and error messages being visible to adversarial users:
- successful attacks get built by attackers stepwise, exploring little-by-little what they can reach / achieve
- giving to attackers as little information as possible is a core rule of secure development. It makes that exploration more difficult. It makes moving laterally more difficult. webERP being open source, attackers already have a lot of info at their disposal
- the exact filesystem path where the app is installed is generally considered a valuable piece of information. It can be used as part of the path _to reach_ remote execution / file-dropping capability
- there are many php.ini settings which the less-savvy sysadmin might turn on, which increase the amount of data visible on screen as part of error messages. Take a look at https://xdebug.org/docs/all_settings#dump_globals as an example of what I mean
- sometimes, even tech savvy sysadmins turn on a dangerous setting by mistake. Or they turn it on temporarily, and forget to turn it off
- I don't see how allowing attackers to see php errors messages by sending hand crafted http requests to openly-accessible urls might help developers, legit users, or sysadmins. There is no benefit whatsoever in allowing that

Last but not least: I have not gone through every php script in /includes with a fine comb, but there is no guarantee that none of those scripts is amenable to carry out data-manipulation when accessed directly from the web. If db access seems unlikely without going through session.php, filesystem access seems to be used much more freely throughout the codebase...

timschofield:
> This situation is why it seems to me that the current and previous crop of webERP developers were not security-minded.
> 

That's interesting. During my involvement with the project we have been audited by a number of security companies, and we have always implemented all their suggestions. I think some of your confusion stems from the fact that you seem to believe that webERP is a website, rather than an application that happens to run in a web browser.

I think I will go with the views of the professional security companies if that is OK with you?

Anyway, enough of this silliness, I have better things to do. I think we are both agreed that the probability of an attack on a webERP installation using this method is either zero, or so close to zero that it makes no difference, so I am now officially moving on, and ending my part in this discussion. The PR has been merged.

PR #715, opened 2025-09-17
Fix direct access to reportwriter page parts

gggeek:
Basically, this adds the same 'prevent direct web access' technique we already use for the installer pages to all the reportwriter pages which are not meant to be accessed directly.

This helps with security, as it avoids the display of php errors/warning messages, which may contain info useful to attackers.

It also helps the automated tests which I am expanding atm.

Also, a comment in index.php and one in 2 apparently unused files.


pakricard:
Hi @gggeek :

I can't see any issue here, but i'm curious why it has been pending to merge since 17th Set.

Is there any issue linked to this PR that we should check?

R

gggeek:
@pakricard this is pending because @timschofield does not agree with the principle of adding such guards against direct access to all php pages but the "public" ones.
I can dig up the relevant discussions if you want...

pakricard:
Thanks @gggeek. PR seem ok to me, but we will wait for @timschofield 's decision,  then.

R

PR #722, opened 2025-09-27
Add more api tests

gggeek:
* remove phpunit warnings for some cases of test failures by moving assertTrue calls at the top
* make api login work: add missing global declarations; add test for it; get default db from config
* make api_session.php more similar to session.php
* make session.php respect more settings from config.php
* make api calls fail with redirect if config.php is missing
* add tests for api calls with the 2 styles of auth


api/api_xml-rpc.php
```diff
@@ -7,4 +7,9 @@
 $dispatchMap = include(__DIR__ . '/includes/api_xml-rpc_definition.php');
 
 $server = new \PhpXmlRpc\Server($dispatchMap, false);
+
+/// @todo we should set this automatically based on $Debug from config.php
```
timschofield, line 11: I believe the debug level for the api shouldn't necessarily be the same as the debug level for the main application

PR #723, opened 2025-09-28
Use utf8mb4 for db, but allow utf8mb3 too

gggeek:
2nd take on the original PR.

This time:
- we default to utf8mb4 for new installs
- we add a config.php setting, so user with existing dbs can keep on using 'utf8' for 3-bytes charsets
- we have the installer query the available charsets at install time (even though all db versions we know we support do have utf8mb4 *)

Fixes an unrelated warning in the installer, too

\* = since it is impossible to test via automation dbs which do not have utf8mb4 support, I tested manually the codepath which makes the installer use utf8mb3, using a mysql 8.3 server

### DIFF
--- a/doc/NEWS.md
+++ b/doc/NEWS.md
@@ -45,4 +45,4 @@ For detailed changelogs of v4 and earlier, please refer to [CHANGELOG.md].
 * introduced file `robots.txt` to avoid accidental indexation of site contents
 * improved the debugging of failed SQL queries when `$Debug` is set to 1
 * css styling changes
-* many bugfixes
+* various bugfixes


config.distrib.php
```diff
@@ -50,6 +50,11 @@
 $DBUser = 'weberp_db_user';
 $DBPassword = 'weberp_db_pwd';
 
+// The DB character set.
+// For mysql/mariadb, utf8mb4 is preferred. If your database version does not support it, you can, depending on the
+// specific version, use the 'utf8' alias for 'utf8mb3'
+$DBCharset = 'utf8mb4';
```
timschofield, line 56: This will break existing installations that won't have this variable in their config.php file. We need an alternative way of doing this.

gggeek: No it won't :-)

- existing users won't have that value set in their `config.php`. That case is handled in session.php and api_session.php, defaulting to `utf8`, which is the same value being used beforehand
- for new installs, the value written in `config.php` is decided by the installer code, which probes for utf8mb4 support

The only users who might get things wrong are users who update, manually copy that line from the new `config.dist.php` into their preexisting `config.php` and do not read the comments nor the upgrade instructions or news file.

PR #728, opened 2025-10-14
First PR for API additions from @jlungo 

dalers:
See [Discussion #720](https://github.com/timschofield/webERP/discussions/720).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,7 +2,6 @@
 /nbproject/
 # As well as Jetbrains stuff
 /.idea/
-*.DS_Store
 # And VSCode
 /.vscode/launch.json
 # And random assorted tooling files


timschofield:
@dalers did you intend to include your .gitignore file in this PR?

Tim

dalers:
Yes, .gitignore should be included. @jlungo added a table tag (I assumed it was intentional and did not question why)

timschofield:
The only change in .gitignore is the addition of *.DS_Store which I maybe wrongly assumed was you because of the DS initials. It doesn't do any harm, but it doesn't seem to add anything either.

dalers:
> The only change in .gitignore is the addition of *.DS_Store which I maybe wrongly assumed was you because of the DS initials.

So far as I'm aware the name and my initials are unrelated. My last commit to .gitignore was Dec. 2024 per the history for .gitignore in GitHub.

image

PR #737, opened 2025-10-21
fix possible double inclusion of functions file from api manual page

### LINKED ISSUE #736: Selecting API Function Reference in webERP Manual results in empty page (again)
Access the API Function Reference from the TOC in the Manual fails again (webERP demo site, localhost on Windows with Xampp and my FreeBSD server). 

I tried to check the on-line static manual (instead of accessing the manual in the demo site), and found it seems to have diverged from the "built-in" manual because there's no "API Function Reference" section (or maybe there never was).

Previously reported in Issue #406 (and fixed and closed).

Image

gggeek:
Should fix #736

### DIFF
--- a/tests/run/BBB_SinglePageTest.php
+++ b/tests/run/BBB_SinglePageTest.php
@@ -50,6 +50,11 @@ class BBB_SinglePageTest extends LoggedInUserTestCase
 				continue;
 			}
 			$pages[] = [$path];
+			// add one extra request for the one manual page which is known to be code-generated
+			/// @todo we could move this to a test dedicated to checking every page within the manual
+			if ($fileName == 'ManualContents.php') {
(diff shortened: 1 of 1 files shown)


gggeek:
As a side note: atm ALL tests are failing for ALL the PRs I sent this evening.

This is not related to changes they introduce - in fact the tests have been failing since Oct 9th, when a change was merged which adds migrations to drop a table from the database but does not update the base db schema to include such change.

The tests fail because they try to log in and browse the site, after having created the database structure. They do not expect a 302 redirection to happen - and this is the case, with the login page redirecting to the 'please upgrade the db' page.

Imho the correct fix is not to make the test suite more accommodating, but rather to make sure that the db created by running the install wizard does not need to execute any updates on start - as that is surely the better user experience. I will create a ticket for that

Issue #733, opened 2025-10-16
PrintWOItemSlip.php not printing the image of the item


pakricard:
Hi:

Latest code (commit #f439938faa28267c1346cf28d16d0cbe4101a15c )

<img width="544" height="425" alt="Image" src="https://github.com/user-attachments/assets/5eb2bcc6-48ce-47d3-8874-15dbfb986c16" />

Does not print the image.

Checking previous versions of the same file:

Version 25/09/2025 (commit #413347ed71e1182afe174be2ec049d145256e5f6 ), the first one using DomPDF:

<img width="569" height="563" alt="Image" src="https://github.com/user-attachments/assets/377a3bed-46b0-4b3e-952a-94eed6f5731a" />

Image is not properly printed.

Version 27/08/2025 (commit #083d50c4639e3dc2a55f2df2d963349de7e45699 ), the last one using TCPDF:

<img width="559" height="617" alt="Image" src="https://github.com/user-attachments/assets/35522e2b-e2fa-42ac-81b5-be57b5e99d11" />

Image was OK.

R

pakricard:
Hi @timschofield:

I just came back to office and tried the latest code, but it does not show the image., just a square.

<img width="582" height="587" alt="Image" src="https://github.com/user-attachments/assets/9b6f6ceb-422a-43b5-8c5d-80fd1c17535f" />

The file is there, and accessible, as the previous script version (the last one using TCPDF) does have access to the image and it shows correctly.

<img width="735" height="571" alt="Image" src="https://github.com/user-attachments/assets/6418f786-bf3d-4615-b452-6cccd49dba22" />

R

P.S.: Nice to have you back. ;-)

timschofield:
Thats strange. Before those changes I was seeing that square, but now with those changes I see:

<img width="608" height="562" alt="Image" src="https://github.com/user-attachments/assets/3a5f54f2-47f5-4125-9dcb-897c28372116" />

timschofield:
> P.S.: Nice to have you back. ;-)

Thanks :)

pakricard:
I'm still seeing the square :-(

Checked SelectProduct.php and to show the image on the screen we use:

```
		$PossibleImageFiles = glob($_SESSION['part_pics_dir'] . '/' . $StockID . '.{png,jpg,jpeg}', GLOB_BRACE);
		if (count($PossibleImageFiles)>0) {
			$ImageFile =  $PossibleImageFiles[0];
		} else {
			$ImageFile ='';
[...]

PR #747, opened 2025-11-02
Proof of concept for a unified file to set domPDF options

pakricard:
Proof of concept for a unified file to set domPDF options instead of individually on every script.
Linked to issue https://github.com/timschofield/webERP/issues/733 

Besides fixing the option of multiple chroot directories, in some scripts (it seems random to me, depending on template used) we have options like:
```
$dompdf->set_option('isHtml5ParserEnabled', true);
$options->set('isHtml5ParserEnabled', true);
$options->set('isRemoteEnabled', true);
```
Probably all these settings would be better if unified in a file, easier to maintain and fewer lines of code.

If this PR gets accepted, all other scripts creating domPDF's will need to be updated to use /includes/SetDomPDFOptions.php

R

### DIFF
--- a/AgedDebtors.php
+++ b/AgedDebtors.php
@@ -442,7 +442,7 @@ if(isset($_POST['PrintPDF']) or isset($_POST['View'])
 		</html>';
 
 	if (isset($_POST['PrintPDF'])) {
-		$DomPDF = new Dompdf($options); // Pass the options object defined in SetDomPDFOptions.php containing common options
+		$DomPDF = new Dompdf($DomPDFOptions); // Pass the options object defined in SetDomPDFOptions.php containing common options
 		$DomPDF->loadHtml($HTML);
 
 		// (Optional) Setup the paper size and orientation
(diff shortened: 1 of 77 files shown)


pakricard:
You have way more experience setting up webERP installations, but my POV is:

If symlinks for image folders is a common thing (I have no clue), setting it up in SystemParameters.php make sense, as most installations will need to set it up.

If it just a few installations using it, then the majority of users will enjoy a simpler setup if it is in config.php. Setting the symlink is not a trivial task sometimes (due to permissions, ownership conflicts, etc), so the average user is probably not going to set this up, but an IT expert. Then for them, update config.php should be an easy task. 

R

timschofield:
Thanks Ricard. I don't have a problem with this in general. Should the variable $SymlinkImageDir set in config.php, or is it better in the config table as a system parameter? I am thinking that as is, the installer will set it automatically to an empty string, but if it is set in SystemParameters.php then it would be part of the standard implementation routine.

Tim

timschofield:
Thanks Ricard, I have merged this. 

Lets see if any one shouts about the config.php/SystemParameters issue and change it if need be.

Thanks
Tim

pakricard:
Thanks, Tim.

PR #775, opened 2025-11-25
Various Code Fixes:

aingelc12ell:
- duplicate switch cases
- simplify conditions
- use match() instead of if|else returning single output
- missing default return value
- simplify dirname() with levels
- merge similar switch cases
- use implode() instead of loop
- remove similar boolean output
- in_array() instead of array_search()
- use of min() instead of if series

### DIFF
--- a/AddCustomerContacts.php
+++ b/AddCustomerContacts.php
@@ -264,7 +264,7 @@ if (!isset($_GET['delete'])) {
 	echo '<field>
 			<label for="ContactNotes">', __('Notes'), '</label>
 			<textarea cols="40" name="ContactNotes" rows="3">',
-				( isset($_POST['ContactNotes']) ? $_POST['ContactNotes'] : '' ),
+				($_POST['ContactNotes'] ?? ''),
 				'</textarea>
 			<fieldhelp>', __('Any notes on this customer contact'), '</fieldhelp>
 		</field>';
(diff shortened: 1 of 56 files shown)


timschofield:
Thanks very much for this. Did you see - https://github.com/timschofield/webERP/discussions/774 - I am wondering how this PR fits in with it?

Tim

aingelc12ell:
PR goes for master;
for confirmation as a bug:
- includes/PDFSalesAnalysis.php line#90 for subtraction operation

timschofield:
Hi, if includes/PDFSalesAnalysis.php is the only bug fix, can you remove it from this PR and do a separate commit for that file with the word "Fix" at the start of the commit message?

I have a script that cherry picks the bug fix commits and adds them to the stable branch as well as master.

Tim

aingelc12ell:
I had the understanding that minor bug fixes are not bugs, or while being tested for any effects on the other parts, hence they are committed to the master. I also had the understanding that the changes that goes to stable branch can come from the master in due time or when they qualify as part of the "bugs".

timschofield:
> I had the understanding that minor bug fixes are not bugs, or while being tested for any effects on the other parts, hence they are committed to the master. I also had the understanding that the changes that goes to stable branch can come from the master in due time or when they qualify as part of the "bugs".

You are correct that everything gets committed to master, but then any bug fix commits are cherry picked into stable. The problem here is that the bug fix is mixed in with other changes, so can't be cherry picked.

Tim

Issue #755, opened 2025-11-10
Z_index.php is listed in the "scripts" table but does not exist


dalers:
I've been poking around the menu system learning how to modify and add menu items when I noticed Z_index.php is included in the scripts table but the file does not exist (although iirc it did once though). Should the entry be removed from the scripts table?

<img width="1382" height="629" alt="Image" src="https://github.com/user-attachments/assets/d9a9a4dd-b7d6-400c-aadd-04d2a6a69e57" />

timschofield:
Z_index.php was initially the menu page for the utilities scripts, but became part of index.php with the utilities module. It was part of Phil's initial upload in Feb 2004.

I think it can be removed from the scripts table.

Tim

dalers:
Resolved by PR https://github.com/timschofield/webERP/commit/46792874b81aa36b2f351bd62e7ad96ad5932ac5

Issue #754, opened 2025-11-10
Sys admin cannot run Z_ImportSuppliers.php


dalers:
The System Administrator cannot run Z_ImportSuppliers.php

Can anyone say how the security is controlled and how to fix this? Z_ImportSuppliers.php is not in the menu, it doesn't seem to be listed in [Main Menu > Setup > Page Security Settings] and there doesn't seem to be any in-script security setting. What am I missing?

Should the Sysadmin Administrator security role in general allow running any script?

I recall either sales or purchasing config required security role Accountant. I assumed the config was considered highly sensitive and therefore only an Accountant was allowed to change. 

Fwiw Z_ImportPartCodes.php could be run with security role System Administrator.

Also, given the other import scripts available in the menu, could Z_ImportSuppliers.php's absence from the menu be an error of omission?

<img width="1082" height="614" alt="Image" src="https://github.com/user-attachments/assets/fdb30f6a-8a19-4e98-b1e0-c90a3feab4d4" />

timschofield:
I can't check from here, but I suspect it hasn't got an entry in the scripts table. Best way is to create a new DB update file containing the function NewScript()

Tim

dalers:
Resolved by PR https://github.com/timschofield/webERP/commit/46792874b81aa36b2f351bd62e7ad96ad5932ac5

PR #795, opened 2025-12-16
menu cleanup

dalers:
- Removes (de-registers) non-existing Z_index.php
- Inserts menu item [Utilities > Maintenance > Import Suppliers from .csv]
- Sets security token for all Z_Import... scripts to the admin (if applicable)

See issues https://github.com/timschofield/webERP/issues/755 and https://github.com/timschofield/webERP/issues/754, and discussion https://github.com/timschofield/webERP/discussions/789

### DIFF
--- a/GetStockImage.php
+++ b/GetStockImage.php
@@ -204,6 +204,13 @@ if ( $automake AND !isset($FileName) ) {
 			$fyh = $fy + $fh - 1;
 			$Textwidth = $fw * mb_strlen($Text);
 			$px = (imagesx($im) - $Textwidth) / 2;
+			// Cast to int to avoid PHP 8.4 deprecation: float to int implicit conversion
+			$Textwidth = (int)$Textwidth;
+			$px = (int)$px;
+			$fy = (int)$fy;
+			$fh = (int)$fh;
+			$fw = (int)$fw;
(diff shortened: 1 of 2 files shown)


dalers:
It struck me that I was using an update not to modify the schema but only to change the configuration of some menu items, which is something many companies may likely do or want to do.

How should a company manage their "custom" menu configuration when the project makes changes? In this case (issue https://github.com/timschofield/webERP/pull/795), if a company has modified their [Utilities > Maintenance ] sub-menu, it seems they will have to re-implement their changes after updating to update 32 (update 32 assuming the PR is accepted).

Anyone have any comments or recommendations? Should the update system be used only for schema changes, and "recommended" or "default" configuration managed seperately in some other way?

To discuss further, please use discussion topic "Custom menu configurations and db update files"  https://github.com/timschofield/webERP/discussions/758#discussioncomment-15272097

dalers:
P.S. sorry for the very messy commit history (including going down a rabbit hole and adding a file that was later abandoned).

I had intended _**not**_ to send the full commit history, just one commit based on the current project master branch and the head of my "ds/menu_cleanup" branch, but I couldn't figure out how to accomplish that exactly....

PR #827, opened 2026-02-02
Audit for scripts

pakricard:
Audit for Script: Records who executed what script, when and for how long did it run. AuditScripts.php shows the results for audit and optimization purposes, similar to AuditTrail.php


dalers:
@pakricard thanks for some nice scripst! :-)

Re AuditScripts.php, how do _**you**_ access it on your production system? Do you add a link to the menu system in some module/category? Do you save the link in your "Commonly used scripts", or do you just remember and type the URL manually? It would seem useful to have a link in the standard menu but I'm curious what works for you. Cheers! :-)

PR #832, opened 2026-02-08
Fix BackupDatabase.php

mrroyh:
- Make BackupDatabase.php work with languages other than English.
- Fix the delete backup path.

### DIFF
--- a/BackupDatabase.php
+++ b/BackupDatabase.php
@@ -14,9 +14,10 @@ if (isset($_GET['BackupFile'])) {
 
 	$BackupFiles = scandir($PathPrefix . 'companies/' . $_SESSION['DatabaseName'], 0);
 	$DeletedFiles = false;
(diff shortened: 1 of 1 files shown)


timschofield:
Thanks for this contribution, I have merged it into master.

Tim

dalers:
@mrroyh Thanks Roy! I have never tried the backup script before but will do so now because you felt it worthwhile to fix and submit a PR.  :-)

Btw, are you also the Roy contributing to OpenWrt? How does webERP fit into your world? There's only a handful of people actively posting in the project repo and we don't hear enough from the everyone else. Cheers!

mrroyh:
Hi @dalers, im just trying weberp out. i think all system must have backup mechanism and i find the backup database script is easy to use but do not work on my language, so i tried a little fix.

yes i add a new device support in openwrt.

PR #845, opened 2026-02-17
Bank adjustments on  automatic currency updates

pakricard:
If using automatic daily exchange rates updates, webERP did not update the value of foreign currencies bank accounts, leading to discrepancies that had to be manually reconciled via "Calculate and Post Exchange Difference" from BankReconciliation.php script.

This PR uses the same logic previously only used in Currencies.php to update the foreign currencies bank accounts.

### DIFF
--- a/includes/LoginFunctions.php
+++ b/includes/LoginFunctions.php
@@ -208,13 +208,9 @@ function userLogin($Name, $Password, $SysAdminEmail = '') {
 							$CurrenciesResult = DB_query("SELECT currabrev, rate FROM currencies");
 							while ($CurrencyRow = DB_fetch_row($CurrenciesResult)){
 								if ($CurrencyRow[0]!=$_SESSION['CompanyRecord']['currencydefault']){
-									$OldRate = $CurrencyRow[1];
 									$NewRate = GetCurrencyRate($CurrencyRow[0],$CurrencyRates);
 									if ($NewRate == '') {
 										$NewRate = 1;
-									}
-									if ($OldRate != $NewRate) {
-										AdjustBankAccountsDueToCurrencyExchangeRate($CurrencyRow[0], $OldRate, $NewRate);
 									}
 									DB_query("UPDATE currencies
 											SET rate='" . $NewRate . "'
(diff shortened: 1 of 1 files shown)


pakricard:
The Checks show one error: 

1) BBB_SinglePageTest::testAccessToAllPages with data set #107 ('/GenerateManualPDF.php')
Symfony\Component\HttpClient\Exception\TimeoutException: Idle timeout reached for "http://localhost/GenerateManualPDF.php". 

but I think it is unrelated to this PR.

timschofield:
Hi Ricard, normal accounting convention is to only revalue bank accounts on a periodic basis, either once a month, or even once per financial year. Revaluing on a daily basis will lead to lots of very messy and confusing transactions in gltrans.

Tim

pakricard:
The problem living and doing  business in IDR, forces us to reevaluate the positions in foreign currency quite often, as the currency swings quite a lot.

Would it make sense to you if I check the date of the last adjustment and then just perform the adjustment only once a month?

R

timschofield:
I think we need to be careful here. We need to be compatible with SSAP 20. At a minimum the user should be able to decide when to do this, normally at end of month or end of financial year.

pakricard:
OK, Tim, I get your point ;-)
Then I see 2 options ahead that could be good for both of us:

1) In SystemParameters.php change the flag
image

to:
- Manually
- Daily
- Monthly
- Yearly

Then, if User selected "Monthly" or "Yearly" then we run the function AdjustBankAccountsDueToCurrencyExchangeRate() at the first month change or year change. If user selected "Manually" or "Daily" then we don't. This should be is easy to code. 

2) I fully remove the calls to AdjustBankAccountsDueToCurrencyExchangeRate() from LoginFunctions() and we keep webERP functionality regarding auto-update as-is. (lines 216-218 and lines 236-238)
```
if ($OldRate != $NewRate) {
    AdjustBankAccountsDueToCurrencyExchangeRate($CurrencyRow[0], $OldRate, $NewRate);
}
```

I think the other changes in the PR code could be kept, and I can always restore the calls to AdjustBankAccountsDueToCurrencyExchangeRate() in my code. 

R

timschofield:
My preference would be option 2. I am nervous of this being done automatically and would prefer it to be left in the control of the organisations accountant when to apply these adjustments. I understand where you're coming from on this, but it upsets my accountant brain :-)

I agree there is other good stuff in it. Not sure why it is still failing the tests, I have changed the tests to work with GenerateManualPDF.php so am hoping when we merge this PR then the failures will go away.

Tim

pakricard:
> My preference would be option 2.

Done!

This last commit tests also passed OK. previous commit the tests failed on timeout.

PR #848, opened 2026-02-17
Currency exchange diff account

pakricard:
Solves https://github.com/timschofield/webERP/discussions/842. 
Uses the functions of UIGeneralFunctions in CompanyPreferences.php to build the fieldset.
Must be merged after https://github.com/timschofield/webERP/pull/845 as it depends in some changes already done there.


timschofield:
Hi @pakricard there appears to still be a call to the AdjustBankAccountsDueToCurrencyExchangeRate() function on line 152 of Currencies.php. Should this be removed?

Tim

pakricard:
The code inside AdjustBankAccountsDueToCurrencyExchangeRate() function was previously in Currencies.php around lines 151-228. 

Now the same code is encapsulated in that function. But functionality of Currencies.php remains the same (except the GL account used that now is the new field $_SESSION['CompanyRecord']['currencyexchangediffact']). 

To my understanding, it is OK. I have been using this updated Currencies.php these last few days and works OK on my side.

R

timschofield:
It appears to me to be automatically posting to gltrans. This should only happen when the user selects to do this from the bank reconciliation.

pakricard:
> It appears to me to be automatically posting to gltrans

Yes. I committed it on 04/06/2013, (almost 13 years ago) ;-), in 59951aac0bf527a3182e02aa0d73e109924e56a4 . We have lost the discussion from the forum, thought http://www.weberp.org/forum/showthread.php?tid=463

After all these years, I assumed it was correct. Should we take this out? Or should we keep it?

R

PR #860, opened 2026-02-27
increase size of stockid and description

dalers:
Increase stockmaster.stockid from varchar(24) to varchar(64).
Increase stockmaster.description from varchar(50) to varchar(255).


dalers:
Thanks Tim. I'll wait for your update 44 to add the stock item notes feature and then submit a PR for update 45 to complete moving "manufacturers" to "brands" (table name, table column names, informal fk relationships, and code references to table columns).

timschofield:
Thanks Dale, I have already committed that change

tim

PR #908, opened 2026-04-17
add skip and update options to Z_ImportStocks.php

dalers:
This PR is to add skip and update capability to Z_ImportStocks.php for when the stockid given in a CSV import file row already exists in the database. The previous behavior was to abort further import and the user would need to edit their CSV file before attempting import again.

The use case for the "update current data" option is for providing additional data related to the same stockid. I.e., when the same stockid is provided in multiple rows for multiple stock item notes, multiple suppliers, multiple approved supplier OEM part numbers, etc. However, only the update checkbox is provided in this PR and the business logic will be added shortly as detailed requirements become understood. The legend for the "update" checkbox indicates the feature has yet to be implemented (applicable rows will be skipped instead - and reported).

### DIFF
--- a/Z_ImportStocks.php
+++ b/Z_ImportStocks.php
@@ -12,11 +12,16 @@ echo '<p class="page_title_text"><img alt="" src="' . $RootPath . '/css/' . $The
 		__('Import Stock Items from .csv') . '" />' . ' ' .
 		__('Import Stock Items from .csv') . '</p>';
 
-// If this script is called with a file object, then the file contents are imported
-// If this script is called with the gettemplate flag, then a template file is served
-// Otherwise, a file upload form is displayed
+// If this script is called with a file object the file contents are imported.
+// If this script is called with the gettemplate flag a template file is served.
+// If neither, a file upload form is displayed.
 
-// The CSV file must be saved in a format like the template in the import module I.E. "RECVALUE","RECVALUE2". The CSV file needs ANSI encoding for the import to work properly.
+// The CSV file must be saved in a format like the template in the import module I.E. "RECVALUE","RECVALUE2".
+
+// The CSV file must be encoded in ANSI for the import to work properly (preserved "old" comment)
+// TODO should this comment be deleted? There is code below to "remove UTF-8 BOM if present"
+
+// stockitemnotes.note is type "text" (65KB = 65K 8-bit characters) but an import row is limited to 10K total bytes.
 
 $FieldHeadings = array(
 	'StockID',         	//  0 'STOCKID',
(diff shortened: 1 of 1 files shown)


dalers:
Hmm, all the CI tests fail unexpectedly "with 3 test failures in the API test suite." I will investigate further until the tests pass but any comments or suggestions would be sincerely appreciated. :-)

I tested the logic before sending the PR but could I have known about the CI test failures?

From GitHub:

I'll help you find a solution for the failing job. Let me start by retrieving the logs for that job.Let me extract the repository information from the PR URL and try again:

# Solution for Job 71909770167
## Problem
The job failed with 3 test failures in the API test suite. All three failures are in `XXX_ApiTest` and share the same root cause:

```
The xmlrpc response has an unexpected fault code
Failed asserting that an array contains 2.
```

The failures occurred in:
1. `testLogin` 
2. `testInCallAuth`
3. `testSessionAuth`

All three tests are failing at the same assertion point in `ApiTestCase.php:83`, indicating the XML-RPC response is returning an unexpected fault code instead of the expected code `2`.

## Root Cause
The API authentication tests expect an XML-RPC fault code of `2`, but the server is returning a different fault code. This suggests either:
- The XML-RPC API response format has changed
- An authentication service is not properly initialized during testing
- The fault code constant or validation logic is incorrect

## Solution

You need to examine and fix the authentication logic in the API test case. Update `tests/src/ApiTestCase.php` at line 83 to handle the actual fault code being returned:

**Step 1:** Check what fault code is actually being returned by adding debugging:

```php
// In ApiTestCase.php around line 83
public function assertValidFaultCode($response) {
    $this->assertIsArray($response);
    // Log the actual fault code for debugging
    if (isset($response['faultCode'])) {
        echo "Actual fault code: " . $response['faultCode'] . "\n";
    }
    // Update this assertion to match the actual fault code being returned
    $this->assertContains($response['faultCode'], [2]); // Change as needed
}
```

**Step 2:** Verify the XML-RPC server is properly configured and returning fault codes correctly. Check if there's a setup issue in the test bootstrap or if authentication credentials are missing.

**Step 3:** Once you identify the correct fault code, update the test expectations accordingly or fix the authentication configuration to return fault code `2`.

The fix likely involves either correcting the expected fault code in the assertion or ensuring the authentication system is properly initialized during the test run.

timschofield:
It looks like the problem was on the github server. I am rerunning the tests and they appear to be succeeding now

Tim

dalers:
> It looks like the problem was on the github server. I am rerunning the tests and they appear to be succeeding now

Woo hoo! :-)   is there any way I could have known that or done anything about it?

timschofield:
> > It looks like the problem was on the github server. I am rerunning the tests and they appear to be succeeding now
> 
> Woo hoo! :-) is there any way I could have known that or done anything about it?

No, looking at the error message it was saying the xmlrpc client couldn't log in to the server, and your commits didn't touch that area, so I just tried running them again and the passed. I have had other tests fail for similar reasons and then pass when I rerun them, so it's always worth a go.

I don't know whether the tests can be run locally, I am sure they canbe but I've never looked into it

PR #935, opened 2026-06-16
Add PN Links to BOMIndented.php

dalers:
Add links from Part Numbers in an Indented Bill of Materials Listing (BOMIndented.php) to SelectProduct.php similar to BOMInquiry.php listing. Discussion https://github.com/timschofield/webERP/discussions/934#discussion-10263974

1. is there a macro or particular structure I should be using to add a link to some string?

2. the links will also appear in a generated PDF, which could be convenient for most users but sometimes night be too much information. It would still be good to support options when generating a PDF but I imagine a PDF could be cleansed after generation if ever needed.

### DIFF
--- a/BOMIndented.php
+++ b/BOMIndented.php
@@ -223,7 +223,7 @@ if (isset($_POST['PrintPDF']) or isset($_POST['View'])) {
 		$Symbol = ($Level > 0) ? '|_ ' : '';
 
 		$HTML .= '<tr class="striped_row">
-					<td>' . $Indent . $Symbol . $MyRow['component'] . '</td>
+					<td>' . $Indent . $Symbol . '<a href="' . $RootPath . '/SelectProduct.php?StockID=' . urlencode($MyRow['component']) . '">' . $MyRow['component'] . '</a>' . '</td>
 					<td>' . $MyRow['mbflag'] . '</td>
 					<td>' . $MyRow['description'] . '</td>
 					<td>' . $MyRow['loccode'] . '</td>


dalers:
3. how would you describe to a user how the links work? Is it correct to say: "The Part Numbers in an Indented BOM (BOMIndented.php) link to viewing the inventory item (SelectProduct.php)" ?

timschofield:
>     2. the links will also appear in a generated PDF, which could be convenient for most users but sometimes night be too much information. It would still be nice to have some options to select from when generating a PDF, but I imagine a PDF could be cleansed if it was ever needed.

Technically it would be possible, but messy. The new reports simply generate the HTML and then the same HTML can be sent to the browser to view it, sent to DomPDF to generate a PDF, or sent to PHPSpreadsheet to generate a spreadsheet. If I understand you correctly we would have to do some preprocessing to the HTML depending on what was being done to it?

Tim

dalers:
> Technically it would be possible, but messy.
>> the links will also appear in a generated PDF

I was assuming the PDF could be edited after creation by some PDF tools suite to delete any and all links (presumably would be simplest). Adobe Acrobat Pro or LibreOffice Draw could be used, maybe a "print to PDF" if it outputs "text only", but also pdftk (the PDF Tookkit) could be used to remove link objects from a PDF.

My feeling is that links in PDFs are OK for now and if it becomes important, the simplest solution may be to require pdftk on the server and use it in the script (adding a config variable to [Setup > System] for Enable/Disable and possibly another for the path to pdftk with command arguments).

P.S. PuMyPDF could be used with a short Python script if only specific links need to be edited.

Issue #953, opened 2026-07-15
Stock item Notes does not render line endings when viewed


dalers:
This issue applies to Stock Item Notes but the same scenario also applies to Customer notes (I'm not going to create an issue now because it's not a hard need at the moment).

Line ends in a stock item note are not rendered when viewed:

<img width="1624" height="562" alt="Image" src="https://github.com/user-attachments/assets/866c7f8e-8485-4fb8-b207-889630d5ebe4" />

even though the text includes line endings as shown in the editor:

<img width="1624" height="562" alt="Image" src="https://github.com/user-attachments/assets/e6aefa33-0fe1-4b37-aaea-e8868c80fd79" />

There are reportedly two basic strategies for this situation:

1. Use the nl2br() built-in php function (newline to break). If the code was echo nl2br($note);, the server would automatically insert `HTML <br>` tags wherever you pressed "Enter," forcing the browser to show the gaps. 

2. Use CSS white-space. Alternatively, the page could use a CSS rule like white-space: pre-wrap; on the note container, which tells the browser to respect the carriage returns exactly as they are stored in the database.

What would you suggest as the solution? Is there text handling elsewhere in webERP that can be the pattern for text handing here?

dalers:
I know just enough about styling and CSS to know I know absolutely nothing about styling and CSS but I see there are two white-space rules in the Professional theme:

```
dale@TMC-Eng2:~/weberp/css/professional$ grep "white-space" *.css
default.css:white-space: nowrap;
main.css:white-space:nowrap;
dale@TMC-Eng2:~/weberp/css/professional$
```

However, not knowing what "nowrap" does or whether it will need to be over-ridden for the stockitem note text box, I've got a solution working using nl2br(). Unless someone can argue the solution should be in CSS, I'll submit a PR for fixing using nl2br().

timschofield:
nl2br() works if you use it in line 114 of StockItemNotes.php and I think that is the best solution.

Tim

timschofield:
Can this be closed now?

dalers:
Issue confirmed resolved.

PR #954, opened 2026-07-16
Render lines correctly when viewing stock item note

dalers:
This PR resolves issue #953 by rendering seperate lines in a stock item note correctly (the lines were previously run together), lines were however shown correctly when editing.

### DIFF
--- a/AddStockItemNotes.php
+++ b/AddStockItemNotes.php
@@ -111,7 +111,7 @@ if (!isset($Id)) {
 		while ($MyRow = DB_fetch_array($Result)) {
 			echo '<tr class="striped_row">
 					<td>', ConvertSQLDate($MyRow['date']), '</td>
-					<td>', $MyRow['note'], '</td>
+					<td>', nl2br(str_replace('\r\n', "\n", $MyRow['note'])), '</td>
 					<td><a href="', htmlspecialchars($_SERVER['PHP_SELF'], ENT_QUOTES, 'UTF-8') . '?Id=', $MyRow['noteid'], '&StockID=', $MyRow['stockid'], '">' .  __('Edit').' </td>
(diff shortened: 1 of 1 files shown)


AddStockItemNotes.php
```diff
@@ -111,7 +111,7 @@
 		while ($MyRow = DB_fetch_array($Result)) {
 			echo '<tr class="striped_row">
 					<td>', ConvertSQLDate($MyRow['date']), '</td>
-					<td>', $MyRow['note'], '</td>
+					<td>', nl2br(str_replace('\r\n', "\n", $MyRow['note'])), '</td>
```
timschofield, line 114: Hi Dale, you don't need the str_replace() here, nl2br() does the job for you:

`<td>', nl2br($MyRow['note']), '</td>`

Tim

dalers: I **_will_** go back and re-test but iirc it was necessary to prevent quote chars in the text from being converted to html (and instead be rendered as written).

dalers: Hi Tim, I agree with you, str_replace() is not needed to render the text in lines and quote chars are unaffected.

Copilot for GitHub recommended this after being re-directed because it's first proposal did produce lines but also changed quote chars to HTML special chars. This was Copilot's 2nd attempt.

I'm curious why the str_replace() was recommended. It seems to me to be doing a UNIX to DOS end-of-line conversion but for what benefit?

I will close this PR un-merged and submit a new PR after sync my fork and removing the str_replace().

dalers:
abandoned, will update and re-submit

AddStockItemNotes.php
timschofield: Hi Dale, you should be able to just add a commit to this PR.

Tim

dalers: Hmmm, thought about that but had never done before and the sledgehammer was handy. Next time......

PR #966, opened 2026-07-22
Force password change flag

pakricard:
Gentle reminder to users to change their passwords after the MaxPasswordAge days.


pakricard:
@timschofield : I am not sure if there is a smarter way to check if the password is up-to-date (not expired) than in header.
Any ideas will be welcomed. ;-)

timschofield:
Hi Ricard, I believe that current thinking is that more frequent password changes leads to weaker passwords as users have to keep remembering new passwords, so will choose something simple.

Should we be looking at a different method of logging in, or is that overkill?

Tim

pakricard:
> Hi Ricard, I believe that current thinking is that more frequent password changes leads to weaker passwords as users have to keep remembering new passwords, so will choose something simple.
> 
> Should we be looking at a different method of logging in, or is that overkill?
> 
> Tim

This came up after I realized some of my users have used the same password set by admin since 2012. 14 years with the same password, not changed even once :-( and I imagined I was not the only one suffering this.

Maybe we can improve the quality of required passwords, as many sites now request (1 Caps letter, 1 number, one symbol, etc) .

I tried 90 days, but I got negative feedback from users, that's why I changed the default to one year, but open to explorre ideas. This PR is just the wake-up call ;-)

R

dalers:
Assuming the system operator wants to enforce users changing their password periodically, this seems a reasonable solution and I would support merging (assuming the default condition is the current behavior, I haven't read the commit).

> This came up after I realized some of my users have used the same password set by admin since 2012. 14 years with the same password, not changed even once :-( and I imagined I was not the only one suffering this. 

But this in itself is not necessarily bad. Security is about justified (and pragmatic) risk mitigation. In all enterprise systems I've used recently, changing passwords every 3 months or so was a requirement and I incremented the _xxx at the end of a common root for that login. Assuming my company was hacked and my customers, suppliers and financials stolen, I suggest ambushing me to steal my login probably wasn't part of it. However, if someone needs to claim that security has been increased and they make the call, so be it. 

Of course, if your business has an extremely high value based on proprietary information remaining secret, and ambushing is a real possibility, then by all means add more layers - perhaps a key-based third-party VPN service (I understand there is some direction on VPNs in Europe but I don't know if or what the effect would be).

> Maybe we can improve the quality of required passwords, as many sites now request (1 Caps letter, 1 number, one symbol, etc) .

Yes, that's another option if someone wants it but I would not want it forced on me given the latest thinking is for longer normal-language passphrases. Requiring periodic password changing with requirements for including specific groups of characters has been in vogue for some time now and there will be business leaders who will demand them. However, cybersecurity standards have shifted away from forcing users to change passwords periodically.

I propose we have a general system that a) allows "the old way" for users and business owners similar to the ones @pakricard describes, but also b) can be configured to claim conformance to NIST SP 800-63B, Microsoft & Cyber National Agencies (NCSC, CISA) and PCI-DSS standards.  

If @pakricard needs these features to meet user expectations, there is no reason why (a) can't come first, so long as its' implementation doesn't preclude (b) or make (b) more difficult.

Of course, we can simply chose to **not** implement periodic password changing or character requirements as they are against current cybersecurity standards and that is that. Do you think you can do that @pakricard? Will you implement forced password changing regardless?

From Gemini:

Historically, changing passwords every 30, 60, or 90 days was the standard practice. However, empirical evidence showed that this requirement actually **lowered** system security. Forced rotation causes users to pick predictable variations (e.g., changing `Spring2025!` to `Summer2025!`), making it easier for attackers to guess new passwords once an older version is leaked.

## 1. Official Cybersecurity Standards

Major cybersecurity frameworks have formally updated their guidance to discourage mandatory periodic password changes:

* **NIST SP 800-63B (Digital Identity Guidelines):** Explicitly states that organizations **should NOT force periodic password expiration**. Instead, password changes should only occur if there is evidence of a breach or compromise.
* **Microsoft & Cyber National Agencies (NCSC, CISA):** Have removed password expiration defaults from Active Directory and enterprise security baselines.
* **PCI-DSS (Payment Card Industry):** PCI DSS v4.0 updated its controls to reflect that if multi-factor authentication (MFA) is implemented, mandatory periodic password changes are no longer required.

### What standard policies recommend instead:

1. **Length over complexity:** Requiring longer passwords or passphrases (e.g., 15+ characters) rather than requiring special characters or numbers.
2. **Breached password screening:** Automatically checking new user passwords against known breach databases (like *Have I Been Pwned*) to reject known compromised strings.
3. **Multi-Factor Authentication (MFA):** Enforcing MFA for account logins, which mitigates the risk of compromised static passwords.

## 2. Industry Contexts Where Periodic Changes *Still* Exist

Despite modern guidelines, forced password rotations remain common due to legacy infrastructure or strict internal compliance requirements:

### Formal Policies (Mandatory Rotation)

* **Legacy Corporate Environments:** Many enterprises still enforce 60- or 90-day Active Directory domain password expiration because their internal IT security policy has not been updated to modern standards.
* **Privileged & Administrative Accounts:** High-privilege infrastructure credentials, database admin keys, and service account API tokens often follow strict, automated password rotation schedules (often 30 to 90 days) via Privileged Access Management (PAM) tools.
* **Compliance Frameworks:** Some regulatory bodies or external auditors still mandate periodic resets as a tick-box requirement.

### Informal Procedures & Best Practices

* **Compromise-Driven Resets:** The modern standard policy for standard accounts—passwords are only changed immediately if an account is compromised, exposed in a third-party data breach, or if a device is lost or stolen.
* **System Offboarding:** Forcing password/credential updates across shared operational systems whenever an employee or contractor with access leaves the organization.

timschofield:
I think my preference is to push the user towards longer and better passwords without any time limits. I have a 21 character password that contains a seemingly random mix of upper and lower case with digits and symbols that I can easily type in. It's fairly secure for most non critical things and I have used it for years on those sort of sites and apps.

pakricard:
Thanks, @timschofield and @dalers for your detailed input. I wasn't aware of the current trend in security, so I coded that was obvious to me.

But I see your point and makes sense.

The main issue I want to fix is that in webERP the combo username/password is created by an admin via WWW_Users.php. The admin later emails the credentials to the user, and the user keeps that password for years to come. User password is known by the admin. Admin might be sending the same default password over and over to everybody. To avoid this, I would like to force the user to change the password.

Maybe instead of password age, we can set a flag “forcepasswordchange” for every user, and when the password is set or updated by admin in WWW_Users, then the flag is set to true. webERP shows the message poking the user to change the password. UserSettings.php sets the flag to false once the user has changed the password.

In case of breach, or wanting a system-wide change of passwords, just resetting all flags to true, would do the trick.

Breached password screening and MFA seem too complex for me now, but, changing to a longer password (length over complexity), would make all the sense.  

Does this option make sense to you?

dalers:
> The main issue I want to fix is that in webERP the combo username/password is created by an admin via WWW_Users.php.

I'm in favor of updating authentication and in particular how the initial password is set to meet current recommendations but imho we should set the bar as low as possible to keep the system simple (while still complying).

> Maybe instead of password age, we can set a flag “forcepasswordchange” for every user, and when the password is set or updated by admin in WWW_Users, then the flag is set to true. webERP shows the message poking the user to change the password. UserSettings.php sets the flag to false once the user has changed the password.

I asked Gemini what the current cybersecurity standards say about creating an initial user password. It seems a unique temporary password must be generated and provided to the user and the user must change the password when they login (and the temporarly password expires after a certain amount of time).  Based on Gemini, your proposal seems to meet basic requirements. 

## Gemini - creating an initial password

While these standards (NIST SP 800-63B, NCSC, CISA and PCI-DSS) discourage arbitrary periodic rotations, they **do prescribe explicit requirements and best practices for how initial (temporary or provisioned) passwords must be created, assigned, and handled.**

The unifying goal across these standards is to prevent **credential interception** (e.g., an admin emailing a plain-text initial password) and to ensure that temporary access credentials **cannot be permanently reused.**

---

## 1. NIST SP 800-63B (Enrollment & Authentication)

NIST addresses initial credentials primarily through its Enrollment/Identity Proofing standard (800-63A) and Authentication standard (800-63B):

* **Forced Change on First Use:** When an initial "memorized secret" (password) or temporary activation code is assigned to a user, the system **MUST** require the user to change it immediately upon initial authentication.
* **Secure Delivery:** Initial credentials must be transmitted via an out-of-band channel or delivered securely (e.g., in person or via an encrypted, single-use link). Sending temporary passwords in cleartext emails or SMS is strictly discouraged.
* **Pre-Enrollment Screening:** The initial password generated for or selected by the user must still pass screening against known compromised, breach-list, or context-specific words (like the company or username).

---

## 2. PCI-DSS v4.0 (Requirement 8.3)

Payment Card Industry standards contain some of the most prescriptive rules regarding initial and reset credentials:

* **Unique First-Time Passwords:** First-time passwords for new users (and reset passwords for existing users) **must be unique to each user**. Generating a blanket standard default (e.g., `Welcome2026!`) for all new hires is explicitly prohibited.
* **Immediate Expiration:** Initial/temporary passwords **must be configured to expire upon first use**, forcing the user to create a new, compliant password immediately.
* **Vendor-Supplied Defaults:** Default credentials provided by software or hardware vendors must be changed immediately upon installation/provisioning.

---

## 3. Microsoft & Cyber Agencies (CISA, NCSC)

Modern agency guidelines and enterprise frameworks emphasize moving away from traditional initial "passwords" entirely in favor of single-use temporary credentials or passwordless onboarding.

### Microsoft Entra ID (Active Directory) Model

* **Temporary Access Pass (TAP):** Microsoft recommends issuing a time-limited, single-use passcode (a TAP) for initial onboarding.
* **Zero Initial Password / Passwordless:** Instead of giving a user an initial static password that they must log in with and change, the TAP allows them to register a passwordless authenticator (like FIDO2 passkeys, Windows Hello, or an Authenticator App) directly during their first session.
* **Force Password Change At Next Sign-In:** If a traditional password is provisioned, Microsoft’s identity controls enforce the `UserMustChangePasswordAtNextLogon` flag, rendering the temporary password invalid after one session.

### CISA & NCSC Guidelines

* **Random Generation:** If an administrator must generate a initial temporary password, it must be randomly generated rather than based on user details (e.g., `FirstName+BirthYear`).
* **Short Expiration Window:** Temporary credentials must have a tight time-to-live (TTL)—typically expiring within a few hours or days—even if the user has not logged in to redeem it yet.

---

## Summary of Core Requirements

| Requirement Area | Standard Guidance across NIST, PCI-DSS & Microsoft |
| --- | --- |
| **Uniqueness** | Must be distinct per individual user; generic defaults are banned. |
| **Lifecycle** | Single-use only; must force an immediate change upon first login. |
| **Expiration (TTL)** | Must auto-expire if unused within a short timeframe (hours/days). |
| **Transmission** | Must not be sent via unencrypted plain text (e.g., standard email). |
| **Modern Trend** | Shift toward **Temporary Access Passes (TAP)** to onboard users directly to MFA/Passwordless. |

dalers:
I created a new issue for the current "new user password" not complying with current cybersecurity direction. https://github.com/timschofield/webERP/issues/967

timschofield:
> Maybe instead of password age, we can set a flag “forcepasswordchange” for every user, and when the password is set or updated by admin in WWW_Users, then the flag is set to true. webERP shows the message poking the user to change the password. UserSettings.php sets the flag to false once the user has changed the password.
> 

I like this option. Setting this flag by default for all new users also stops the issue of the admin knowing user passwords.

Tim

pakricard:
> > Maybe instead of password age, we can set a flag “forcepasswordchange” for every user, and when the password is set or updated by admin in WWW_Users, then the flag is set to true. webERP shows the message poking the user to change the password. UserSettings.php sets the flag to false once the user has changed the password.
> 
> I like this option. Setting this flag by default for all new users also stops the issue of the admin knowing user passwords.
> 
> Tim

OK, then. I will code it in the coming days and update this PR. Then, on a second step we can decide on required password complexity or lenght,

R

pakricard:
I think it is done now. Flag is coded, and I encapsulated the password quality verifications in one function, so we can start working on it. As a 1st step I added the config variable PasswordMinLenght so every admin can choose their desired minimum password length. 
Next improvements could probably be forced password complexity (include Caps, or numbers or symbols)

timschofield (approved):

dalers:
After updating my db (2 updates), I was notified that my password had to be updated.

image

I was expecting the changes to have no effect until the admin enables a feature. What is the expected behavior?

pakricard:
@dalers, it is working on my system, so let's try to debug...

In your www_users table, do all users have the flag forcepasswordchange set to zero?
Because that warning message is only triggered at the header (called by all scripts) by:

```
if (isset($_SESSION['UserID'])
    and isset($_SESSION['ForcePasswordChange'])
    and $_SESSION['ForcePasswordChange'] == 1) {
        prnMsg(__('You are required to change your password in User Settings'), 'warn');

```
 and 68.php should have done it at
$SQL = "UPDATE www_users SET forcepasswordchange = 0";

dalers:
@pakricard I don't have access to the database I was commenting on earlier (it's at work behind a firewall and I'm now at my home) but I just upgraded two more databases and I'm not getting notices to change my password.

I'll check the forcepasswordchange flag on the first database on Monday.

Fwiw at the time I was seeing the "you need to reset your password" notices I was trying to set my user timeout to 600 minutes (i.e. 10 hours), except I didn't realize the timeout is stored as a tinyint(4) (max value 125), and entering 600 caused an SQL error. Perhaps this caused the script to behave unexpectedly and resulted in the password issue.

dalers:
> just upgraded two more databases and I'm not getting notices to change my password.

... and the forcepaaawordchange flag was set to zero for all users

dalers:
@timschofield @pakricard I happened to be reviewing my WackoWiki site configuration and thought it might be useful to see the password config in a different app for comparison.

image

Issue #968, opened 2026-07-23
update 67.php fails because hrcolleaguefeedback table could not be created


dalers:
Update 67 fails for me on WSL Ubuntu (Apache: 2.4.58, PHP: 8.3.6, MariaDB: 10.11.14-MariaDB).

<img width="936" height="455" alt="Image" src="https://github.com/user-attachments/assets/ebc7a0b2-ad6e-459b-9f4f-373a7a1a6cfe" />

```
The table hrcolleaguefeedback could not be created
The constraint fk_hrcolleaguefeedback_fromemployee could not be added
ALTER TABLE hrcolleaguefeedback ADD CONSTRAINT fk_hrcolleaguefeedback_fromemployee FOREIGN KEY (fromemployeeid) REFERENCES hremployees (employeeid)
The constraint fk_hrcolleaguefeedback_aboutemployee could not be added
ALTER TABLE hrcolleaguefeedback ADD CONSTRAINT fk_hrcolleaguefeedback_aboutemployee FOREIGN KEY (aboutemployeeid) REFERENCES hremployees (employeeid)
The constraint fk_hrcolleaguefeedback_createdby could not be added
ALTER TABLE hrcolleaguefeedback ADD CONSTRAINT fk_hrcolleaguefeedback_createdby FOREIGN KEY (createdbyid) REFERENCES hremployees (employeeid)
The table hrfeedbackcriteria could not be created
The table hrfeedbackcriteriascores could not be created
```

Six errors in Apache error log /var/log/apache2/error.log

```
[Thu Jul 23 09:51:22.778265 2026] [php:warn] [pid 31308] [client 172.22.192.1:49915] PHP Warning:  mysqli_query(): (HY000/1302): Conflicting declarations: 'COLLATE utf8mb4_unicode_ci' and 'COLLATE utf8mb4_general_ci' in /home/dale/weberp/includes/ConnectDB_mariadb.php on line 63, referer: http://tmc-eng2.ad.telematic.ca/weberp/Z_UpgradeDatabase.php
[Thu Jul 23 09:51:22.780101 2026] [php:warn] [pid 31308] [client 172.22.192.1:49915] PHP Warning:  mysqli_query(): (42S02/1146): Table 'weberp_kayden.hrcolleaguefeedback' doesn't exist in /home/dale/weberp/includes/ConnectDB_mariadb.php on line 63, referer: http://tmc-eng2.ad.telematic.ca/weberp/Z_UpgradeDatabase.php
[...]

PR #969, opened 2026-07-23
Update67

dalers:
Remove collation specification from update


dalers:
works for me, resolves https://github.com/timschofield/webERP/issues/968

Issue #975, opened 2026-07-27
AI agent instructions are not included in repo (originally DATETIME vs TIMESTAMP)


pakricard:
Hi:

During the last few months, we added quite a few columns type TIMESTAMP to record an instant in time when someone created or modified something. But we have some columns DATETIME for similar purposes (audittrail and auditscripts come to mind).

So, should we use one or the other? Having two types for the same concept can cause inconsistencies, so I guess we should move towards one or the other. 

Checked with Gemini to get a quick idea:

<html>
<body>
<html><head></head><body><h3 data-path-to-node="0" style="font-family: &quot;Google Sans&quot;, sans-serif !important; line-height: 1.15 !important; margin-top: 0px !important;"><b data-path-to-node="0" data-index-in-node="0" style="font-family: &quot;Google Sans Text&quot;, sans-serif !important; line-height: 1.15 !important; margin-top: 0px !important;">Core Differences</b></h3><p data-path-to-node="1" style="font-family: &quot;Google Sans Text&quot;, sans-serif !important; line-height: 1.15 !important; margin-top: 0px !important;">While both <b data-path-to-node="1" data-index-in-node="11" style="font-family: &quot;Google Sans Text&quot;, sans-serif !important; line-height: 1.15 !important; margin-top: 0px !important;">DATETIME</b> and <b data-path-to-node="1" data-index-in-node="24" style="font-family: &quot;Google Sans Text&quot;, sans-serif !important; line-height: 1.15 !important; margin-top: 0px !important;">TIMESTAMP</b> store date and time information in MariaDB, they differ fundamentally in <b data-path-to-node="1" data-index-in-node="107" style="font-family: &quot;Google Sans Text&quot;, sans-serif !important; line-height: 1.15 !important; margin-top: 0px !important;">timezone handling</b>, <b data-path-to-node="1" data-index-in-node="126" style="font-family: &quot;Google Sans Text&quot;, sans-serif !important; line-height: 1.15 !important; margin-top: 0px !important;">storage size</b>, <b data-path-to-node="1" data-index-in-node="140" style="font-family: &quot;Google Sans Text&quot;, sans-serif !important; line-height: 1.15 !important; margin-top: 0px !important;">range</b>, and <b data-path-to-node="1" data-index-in-node="151" style="font-family: &quot;Google Sans Text&quot;, sans-serif !important; line-height: 1.15 !important; margin-top: 0px !important;">auto-update behavior</b>.</p>

[...]

PR #979, opened 2026-08-04
Changing TIMESTAMP into DATETIME

pakricard:
Code related to https://github.com/timschofield/webERP/issues/975


timschofield:
MariaDB doesn't support a default of CURRENT_TIMESTAMP for type DATETIME in version 5.5 which is why one of the tests is failing for this PR.

pakricard:
I committed a fix to avoid these changes in 70.php and failed again.

Frustrated, I asked Copilot to fix the issue and it committed 

It does not complain now on 70.php, but just updated one install sql file, not all of them. i simply can't understand why,just one file, but keeps failing.

Copied the Copilot "Explain the error": 

Root cause is in the installer SQL schema, not PHPUnit itself.

### What failed
Job **91948824310** failed in `InstallerTest.php:157` because installation could not create tables.  
The captured failing page shows:

- `Invalid default value for 'logintime'`
- `Invalid default value for 'createdate'`
- then downstream table-missing errors (`departments`, `custcontacts`, `custnotes`) because table creation aborted early.

This is consistent with MariaDB/MySQL strict mode rejecting legacy zero-date defaults like `0000-00-00 00:00:00` or invalid `DATETIME/TIMESTAMP` defaults.

### Fix to apply
Update the install schema (the SQL file(s) used by installer) so datetime/timestamp columns use valid defaults.

Typical changes:

```sql
-- bad (strict-mode incompatible)
logintime DATETIME NOT NULL DEFAULT '0000-00-00 00:00:00',
createdate DATETIME NOT NULL DEFAULT '0000-00-00 00:00:00',

-- good options
logintime DATETIME NULL DEFAULT NULL,
createdate DATETIME NULL DEFAULT NULL
```

or, if semantics require automatic creation time:

```sql
createdate DATETIME NOT NULL DEFAULT CURRENT_TIMESTAMP
```

For “last login” fields, `NULL DEFAULT NULL` is usually safest.

### Why this solves it
Once those invalid defaults are removed, table creation succeeds, and the later “table does not exist” warnings disappear automatically because they were secondary failures.

### Concrete patch pattern
In the SQL DDL used during install (likely your core install `.sql` script), replace any:

- `DEFAULT '0000-00-00'`
- `DEFAULT '0000-00-00 00:00:00'`
- invalid defaults on `DATETIME/TIMESTAMP`

with `DEFAULT NULL` (and allow NULL), or `CURRENT_TIMESTAMP` where appropriate.

### Extra safeguard (optional)
If you need backward compatibility, you can make installer-generated SQL conditional by server version/mode, but the best long-term fix is valid SQL defaults everywhere.

So, I guess the best solution to pass the test is to have different install/sql depending on the server DB. But this is way over my capacity (I never worked on the installer). 

@timschofield Could you please take a look? Or should we just keep the old install/sql files and just apply the change at 70.php?

dalers:
> So, I guess the best solution to pass the test is to have different install/sql depending on the server DB. But this is way over my capacity (I never worked on the installer).

Would that make the installer follow the same pattern as the rest of webERP? I.e. the equivalent of seperate "ConnectDBxxx" files?

> keep the old install/sql files and just apply the change at 70.php?

Does this provide an immediate path forwards and install/sql files can be dealt with seperately? To me this seems the most practical strategy.

timschofield:
Hi @dalers this isn't a MariaDB v MySQL issue, it is specific to certain older versions of MariaDB. Modern versions work fine.

I think there are two easy options:

1 Decide to stop catering for MariaDB v5.5 or
2 Switch to using TIMESTAMP as MariaDB v5.5 handles that in the same way as other versions as do all versions of mysql
3 Adapt the installer to use TIMESTAMP when DB using MariaDB version 5.5

My preference is for 2

timschofield:
> 3 Adapt the installer to use TIMESTAMP when DB using MariaDB version 5.5

We already do this in the installer with stored procedures which were dealt with differently in MariaDB 5.5

pakricard:
> Hi @dalers this isn't a MariaDB v MySQL issue, it is specific to certain older versions of MariaDB. Modern versions work fine.
> 
> I think there are two easy options:
> 
> 1 Decide to stop catering for MariaDB v5.5 or 2 Switch to using TIMESTAMP as MariaDB v5.5 handles that in the same way as other versions as do all versions of mysql 3 Adapt the installer to use TIMESTAMP when DB using MariaDB version 5.5
> 
> My preference is for 2

My preference is for 1). Why? Because at some point we need to let old versions go and don't overcomplicate our code for old stuff. 

Yes, some people might be using MariaDB 5.5, the same way some people was using PHP 5 and now we force PHP 8. They had to adapt, as users of most Open Source projects do.

To me, consistency and code quality is far more important than maintaining compatibility with old versions.

_MariaDB 5.5 was generally released on April 11, 2012, with its final maintenance version (5.5.68) rolling out on May 12, 2020. End of Life (EOL): April 2020 Final Version (5.5.68): May 12, 2020._ 

Even MariaDB stopped maintaining it 6 years ago.

Users who can't (or don't want) move to newer MariaDB versions can stick to webERP 4.15.2, the same way as users with PHP versions older than 8.0. If they want to enjoy all the beauties of webERP 5.0, they must upgrade their PHP and MariaDB. 

We are a tiny group of developers, with limited resources. I don't think it is wise to use more of our time maintaining compatibility with such old versions. 

R

timschofield:
OK, makes sense, I will try to remove 5.5 support this morning.

timschofield:
I have removed support for MariaDB 5.5 and all tests pass now, so am merging this PR

PR #980, opened 2026-08-05
Plm1 - some cleanup and handy links

dalers:
I think this includes everything, I think (hope?) everyone will finds these changes useful.

- show item notes after selecting product
- add BOM-related sub-menu navigation links for selected product
- add navigation link from BOM to Item
- simplify selected item link to WackoWiki (simply "Documents, Drawings and Files" instead of the posh, lengthy and less crisp "Wiki Product Knowledge Base")
- standardize the format of tables that have a row "Action" column at the right
- add Action items "Goto BOM" and "Goto Item" in a Where Used inquiry (previously only linked to a Costed BOM and without indicating it to users in the link text).

2026-08-05 13_12_12 SelectProduct php added BOM-related sub-menu items (and Item Notes)

image

image

I anticipate replacing "Wiki Product Knowledge Base" with "Documents, Drawings and Files" may invoke discussion. For my users, "Notes" are in webERP and are used to share information between users while the wiki interface ("Documents, Drawings and Files") is used for controlled files (e.g. PDFs, ZIPs, Word Docs, Spreadsheets, etc.). Imho this creates a clean separation of function (made possible by the recent addition of Item Notes) and is simple for users to understand. The Wiki provides a simple document management system when an approval workflow isn't necessary but still provides traceability to the user that provided the file with a timestamp, 

If everyone agrees the more descriptive "Documents, Drawings and Files" is "better" being more intuitive for the user, the other supported wikis (DocuWiki and MediaWiki) will need to be updated as well as adding Notes to Suppliers for consistentency.

However, perhaps not everyone will want to name the Wiki interface  "Documents, Drawings and Files", and configuring the link text in [ Settings > System > Wiki Interface ] may be preferable.


pakricard:
Thanks, @dalers, seems like a nice improvement.

> However, perhaps not everyone will want to name the Wiki interface "Documents, Drawings and Files", and configuring the link text in [ Settings > System > Wiki Interface ] may be preferable.

Yes, agree with the configurable text, as for some users "drawings" makes no-sense, but "Extra information" would make more sense to the users.

dalers:
> agree with the configurable text (wiki link text)

I'll see if i can figure out something simple otherwise will update the PR and leave the wiki link text as-is for now.

timschofield:
On a wider note (perhaps I should take my own advice and start a new issue) is a link to a wiki the best way to store this info? Phil is a big fan of Wikis and I understand this, but is it not better just to have a means of storing assorted files such as PDFs or MDs?

dalers:
> On a wider note (perhaps I should take my own advice and start a new issue) is a link to a wiki the best way to store this info?

I guess that would depend on what "better" means to you ;-)

I agree with you @timschofield that a discussion on the "best" way to provide content management functionality (which a wiki provides and which electronic document management is a subset of) should have its own discussion topic and I happen to have created one just now for that purpose: https://github.com/timschofield/webERP/discussions/981

I will remove renaming the wiki link text in  includes/MiscFunctions.php so this PR can (hopefully) be accepted and merged for its other features.

dalers:
The change of wiki link text in includes/MiscFunctions.php has been undone

timschofield:
I have merged this PR

Issue #977, opened 2026-07-28
Accounting errors in SupplierBalsAtPeriodEnd.php for suppliers in foreign currencies


pakricard:
Hi:

We never used SupplierBalsAtPeriodEnd.php until today and I realized that for foreign currencies suppliers, the balance in FX is correct, but the equivalent in home currency (in my case IDR) is wrong. 

<img width="438" height="612" alt="Image" src="https://github.com/user-attachments/assets/c094a051-bbdc-42ab-a9a8-98b7a23f3d43" />

Obviously 0 USD or 0 THB should be 0 IDR, not any random number.

Checked the source code and data in supptrans and I found 2 issues:

1) Some of the data in supptrans (old, probably pre-2024) has rate = 1 / rate , so when calculating the SUM in

`SUM((supptrans.balance)/supptrans.rate) AS balance,
`
gets monstrously wrong (as per example 1 USD = 17.500 IDR or 1 IDR = 0.0000571).

I have not being able to find the exact commit causing this, but must be some related to any INSERT INTO supptrans. My supptrans table has over 14.000 rows, so I need to think of an autoamtic way to find the offernding rates, manually makes no sense.

2) Conceptually, should we use the rate at the time of transaction or today's rate?

As I see it, if the company owes 3.000 XXX to a supplier, it owes to the value of today's home currency, whenever the debt was acquired.

On the other hand, if the company keeps paying its debt to suppliers, the debt in supplier's currency can be zero, but the small differences in exchange rates during the transactions and years keep acumulating.

Seems clear to me that 0 XXX should always be 0 home currencies...

Then:
a) should we abandon the translation of foreign suppliers to home currency and return just the value in supplier's currency?
b) should we keep both (foreign and home) but using today's rate? Will be right today but wrong for past periods
c) should we keep it as is? (will be wrong, for the reasons exposed in my point 2) above

My vote goes to solution a)

R

P.S.: I can imagine @timschofield accountant brain finding 10 reasons why this is wrong ;-)

timschofield:
I haven't looked at this script in many years, not sure I have ever actually used it in a production environment. Phil appears to have done some work on exchange rates in this script back in 2011, but as far as I can see that was the last time anything got changed in the currency logic.

I agree with @pakricard that we should simplify it and just use the suppliers currency. I can't really imagine circumstances where it would be useful to see the balances in the functional currency.

Tim

pakricard:
Done and committed.

R

PR #998, opened 2026-08-26
Standardize navigating from item to bom

### LINKED ISSUE #997: Link from Item to BOM does not use default BOMs.php entry
I added this link in the first place so I'm hoping no one will object to my tweaking it a bit. 

Following the link from an item to its' BOM, a user is presented with the complete hierarchical BOM including expanded sub-assemblies, and not just the top level as they might have wanted.  I propose changing the link to enter the default search entry point in BOMs.php, and which as a level selector.

Currently, following the link from an item to its' BOM the complete hierarchical BOM.

Image

Image

To see just the top level, the user has to use the default BOMs.php entry via the [ Manufacturing > Maintenance > Bill of Material ] menu gives a list of search results and provides an "all" or "top-level-only" selector. I propose to make the link from a selected item behave the same way (and probably change the field description for clarity).

Image

A seperate issue is that navigating between top-level and complete BOMs requires going through a search and the selected BOM is forgotten. To resolve this, and to ease navigation in general, it would be good if a selected BOM was remembered in the same way a selected Item is remembered, and the stock code was pre-filled with the current item / parent. 

Iiuc the the BOM parentid and the Stocks stockid have a 1:1 relationship making them essentially the same but it seems SelectProduct.php/Stocks.php and BOMs.php use different strategies to remember their respective selected Item / Parent. Standardizing on a common approach would provide a more consistent user experience as well as benefit future development but the consolidation would need to be well thought out. If anyone has thought about this before and has ideas please post.

dalers:
Resolves #997 

Includes some heading text changes for clarity, use standard search entry to BOMs.php so level selector can be used, change "Show Levels" selector text for clarity, preserve selected BOM for searching again, remember if BOM top-level or all-levels selected and preserve if BOM is re-sequenced.

### DIFF
--- a/BOMs.php
+++ b/BOMs.php
@@ -198,6 +198,12 @@ if (isset($_GET['SelectedParent'])) {
 	$SelectedParent = $_POST['SelectedParent'];
 }
 
+if (!isset($SelectedParent) and isset($_GET['StockCode'])) {
+	$_POST['Keywords'] = '';
+	$_POST['StockCode'] = trim(mb_strtoupper($_GET['StockCode']));
+	$_POST['Search'] = 'Search';
+}
+
 if (isset($_GET['ShowAllLevels'])) {
 	$_POST['ShowAllLevels'] = $_GET['ShowAllLevels'];
 }
@@ -1126,6 +1132,12 @@ if (isset($SelectedParent)) { //Parent Stock Item selected so display BOM or edi
 
 } //end of if search
 if (!isset($SelectedParent)) {
+	if (!isset($_POST['Keywords'])) {
+		$_POST['Keywords'] = '';
+	}
+	if (!isset($_POST['StockCode'])) {
+		$_POST['StockCode'] = '';
+	}
 
 	echo '<p class="page_title_text">
 			<img src="', $RootPath, '/css/', $_SESSION['Theme'], '/images/magnifier.png" title="', __('Search') , '" alt="" />', ' ', $Title, '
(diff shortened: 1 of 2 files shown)


dalers:
Hi @timschofield, a very illuminating Copilot review!

Iiuc Copilot found the PR correct and reasonable although it found potential issues with the original code.

I'm not able to understand the full impact of the potential issues reported by Copilot, but since the PR resolves the issue #977 and doesn't create any new issues, I think the PR should be merged as-is and if the potential issues reported by Copilot are felt to be significant by someone who understands them better than me, a seperate issue should be created to resolve them. How do you feel about this?

dalers:
Thanks @timschofield for merging. What are your thoughts on the review by Copilot? It's also concerning to me why the regression tests keep failing. The tests seem to pass after restarting (or at least iirc the times I tried they did), but I would have more confidence in them if I knew why they failed.

timschofield:
> Thanks @timschofield for merging. What are your thoughts on the review by Copilot? It's also concerning to me why the regression tests keep failing. The tests seem to pass after restarting (or at least iirc the times I tried they did), but I would have more confidence in them if I knew why they failed.

I have read that Microsoft have overstretched the resources at GitHub by over integrating copilot, causing infrastructure problems. Sometimes the test fail for legitimate reasons though so they always need investigating. 

That is why there has been such a big increase in cost of copilot, they are trying to restrict usage.

PR #1000, opened 2026-08-31
Also show pagination controls on page bottom

dalers:
As a convenience to the user, also show pagination controls at the bottom of the search results (not only at the top).

I have "Maximum Number of Records to Display" in my personal profile to 50. After I've scrolled down the search results and find I need the next page, it would be convenient if there was a "Next" button at the bottom of the page.

Hopefully this will not cause an issue for those who set "Maximum Number of Records to Display" to a small number (e.g. 20, which is what @pakricard uses iirc).


dalers:
Forgot to add screenshot to show finished result.

image

PR #1003, opened 2026-09-03
preserve existing comment when editing BOM line

dalers:
It was intended that when editing a BOM line item the comment box is to be pre-filled with the current comment. However, it does not get pre-filled due to a code mismatch between "comment" and "remark".

### DIFF
--- a/BOMs.php
+++ b/BOMs.php
@@ -404,7 +404,7 @@ if (isset($_GET['Add']) or isset($_GET['Edit'])) {
 						workcentreadded,
 						quantity,
 						autoissue,
-						remark
+						remark AS comment
 					FROM bom
 					INNER JOIN locationusers
 						ON locationusers.loccode=bom.loccode


dalers:
I look at all the failed tests and wonder WHY?!? At least the last commit for which I re-ran the failed tests, the tests still did not complete. Trying again here....

dalers:
All tests completed successfully :-)

timschofield:
Well for instance one of the PRs you submitted earlier was failing. Looking at the error logs what was happening was it was testing the API, but the server for one side was temporarily down when it made the API request causing the test to fail. Not the fault of our code, and when I tried again it was up and the test passed. Like I said it's an MS problem, I know some projects have given up and moved to other platforms. I don't feel we are there yet.

timschofield:
I have seen projects going to Codeberg for instance so I created  https://codeberg.org/TimSchofield/webERP.git/ just in case.

Tim

dalers:
> created https://codeberg.org/TimSchofield/webERP.git/ just in case.

"just in case" of what? MS imploding and GitHub completely collapsing? Iiuc, projects are moving to Codeberg because they identify more with the Codeberg ethos than M$.

For the webERP project, my recommendation is to not spend any effort on migrating unless you feel in your soul that Codeberg is your people and you (and the rest of the community) are willing to give up the free features GitHub offers and create workarounds or alternate workflows instead. I very much would not want to abandon Copilot for Github or make using it more complex (at least not in the next year).

I think there'd be far more benefit from "someone" configuring a simple CI test workflow on a raspberry pi - check out the master branch every night, run some tests, if the tests all pass merge "master" into "master-master" otherwise send emails. This extra layer would be "similar" to the Linux kernel project build structure. Mind you, I've never created such a system and although it doesn't seem would be that difficult if kept simple, building even a simple system would still take time away from doing something else.

We should avoid tying the project to a commercial CI infrastructure. Basically we got hooked by the Github "free" and now we're finding "free" has some issues. To some degree I think Codeberg will have it's own issues.

timschofield:
I said:

> I know some projects have given up and moved to other platforms. I don't feel we are there yet.

Then I said:

> I have seen projects going to Codeberg for instance so I created [codeberg.org/TimSchofield/webERP.git](https://codeberg.org/TimSchofield/webERP.git/) just in case.

Meaning if we ever got to the stage where GitHub's failings got to to the point where we did feel it was time, then I had already set it up on Codeberg. I'm not spending time migrating anywhere (though it is my time to do with what I please), it was a few minutes to create it (some months ago) and is kept up to date by a cron job, periodically pushing the latest code to codeberg.

I have friends in other projects, and GitHub's failings (such as tests failing because of GitHub infrastructure problems) are a regular topic of conversation and some are leaving. As I said, my personal opinion is that we are nowhere near that point.

PR #1006, opened 2026-09-05
add sub-bom folding to BOMIndented.php

dalers:
This PR adds sub-BOM-folding to BOMIndented.php using a new js script. The control is only relevent to the View option (and not PDF). Fwiw, I consider BOMIndented.php as the BOM viewer for everyone to use while BOMs.php is the BOM editor for only those with BOM create/edit privileges.

image

image

The control is visible with a single-level BOM although in this scenario it has no practical value. 

image

### DIFF
--- a/BOMIndented.php
+++ b/BOMIndented.php
@@ -183,9 +183,10 @@ if (isset($_POST['PrintPDF']) or isset($_POST['View'])) {
 					' . __('Indented BOM Listing For') . ' ' . mb_strtoupper($_POST['Part']) . '<br />
 					' . __('Printed') . ': ' . date($_SESSION['DefaultDateFormat']) . '<br />
 				</div>
-				<table>
+				<table id="IndentedBOM">
 					<thead>
 						<tr>
+							<th></th>
 							<th class="SortedColumn">' . __('Part Number') . '</th>
 							<th class="SortedColumn">' . __('M/B') . '</th>
 							<th class="SortedColumn">' . __('Description') . '</th>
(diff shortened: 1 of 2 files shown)


timschofield:
I think that this can be achieved more easily and without the need for JavaScript by using `<details>` and `<summary>` tags

https://www.w3schools.com/tags/tag_details.asp

Tim

dalers:
Hi Tim, thanks for reviewing the PR. I understand a webERP guideline is to use native-HTML if possible to avoid having to fetch and execute js but I'm not an HTML expert so I queried Gemini for its opinion. It seems in this case an HTML-only solution may require considerable refactoring and possibly result in more complex code. 

Gemini reported details tags cannot be seamlessly dropped around standard table row groups without breaking HTML specification rules because webERP tables use tabular layouts. 

It seems there are two issues:

> 1. Table Structure Violations: According to W3C specifications, a table can only directly contain table-related elements. Placing a details tag around multiple rows is invalid HTML and will cause rendering issues across modern browsers:
> 
> 2. CSS Display Flex/Grid Overrides: To use a detail tag inside a table natively, you have to force the component rows to render inside a single structural cell or alter the webERP layout style away from legacy tables and towards modern CSS Grid layouts.

and successfully replacing the js toggle behavior with a details tag will require using the nested single-cell wrapper technique.

You're the expert here TIm but it seems the risk of introducing new issues during refactoring as well as potentially increasing code complexity (if I understand the refactoring implications correctly) may outweigh the benefits of an HTML-only solution. Also note the js is small and only fetched if the user runs the BOMIndented.php script.

Are there other reasons to prefer an HTML-only solution?

Gemini analysis: https://share.google/aimode/VZk62QIzGRXF5bi4f

timschofield:
I would respectfully disagree with Gemini here. This is an example table that uses `<details>` and `<summary>` to achieve this folding. As you can see the additional HTML needed to be added to the table is minimal,

```
<table border="1">
  <thead>
    <tr>
      <th>Top</th>
      <th>Level</th>
    </tr>
  </thead>
  <tbody>
    <tr>
      <td colspan="2">
        <details>
          <summary>
            Click to expand sub-levels
          </summary>
          <table border="1">
            <tr>
              <td>Sub</td>
              <td>Level 1</td>
            </tr>
            <tr>
              <td>Sub</td>
              <td>Level 2</td>
            </tr>
          </table>
        </details>
      </td>
    </tr>
  </tbody>
</table>
```

PR #1015, opened 2026-09-15
Cleanup user settings

### LINKED ISSUE #1013: WWW_Users and UserSettings are not aligned
Some of the parameters in the menu item [ Setup > General > Users Maintenance ] are duplicated in the User Profile, which I would contest leads to user confusion over which takes precidence and where a particular parameter should be set (I said "user confusion", there is no developer confusion ;-) ) 

Imho if functionality can be crisply segregated it should be so that users only have to learn one way (recognising the converse that if functions can't be crisply segrated, duplication will likely be needed e.g. duplicated sub-menu items in "Manufacturing" and "Inventory" modules).  

I also propose moving the desired report (actually PDF) page size from Users Maintenance to the user profile, which will nicely leave Users Maintenance with only settings related to user control and security. 

Image

dalers:
Added missing parameters only in UserSettings.php to WWW_Users.php. Added paper size to UserSettings.php.

Resolves #1013


timschofield (commented):
Would it be more efficient to store the page sizes in an array, and then create the options by cycling through the array?

